PR #3, opened 2016-02-17
update_page should not require parent_id

nickvlku:
The `update_page` method in confluence should not require a `parent_id`.  It looks like the `examples/confluence-update-page.py` also isn't expecting a required parent_id.

I've modified the method to work with a parent_id for old code, and not require it for new code (why I didn't move the named param with default value to the end).

Thanks!
Nick


MattAgile:
Thank you very much! :)

PR #25, opened 2018-02-07
Very basic Crowd client

cattz:


atlassian/VERSION
```diff
@@ -1 +1 @@
-1.0.1
+0.0.0
```
AstroMatt: New version 1.1.0

PR #31, opened 2018-02-19
Feature/bamboo improvements

cattz:
More improvements to the Bamboo API.

:warning: These changes will break backwards compatibility of the Bamboo module!!!

I've abstracted dealing with paginated results, by returning a generator.
This means we can, for example, iterate through all build results fort a given plan, without having to deal with pagination regardless the number of results:

    for result in bamboo.latest_results():
        pprint(result)

Haven't tested this extensively, so worth reviewing and maybe testing it a bit more before merging


AstroMatt:
This looks amazing!
I'll check this out

PR #34, opened 2018-03-30
Add support for parent_id set as Null

svyotov:
The documentation does not work as `parent_id` cannot be set as null.
This changes the method signature, so I do not know if that is fine with you.

```
status = confluence.create_page(
    space='DEMO',
    title='This is the title',
    body='This is the body. You can use <strong>HTML tags</strong>!')
```
This PR allows for `parent_id` being `None`.

Also `self.get('/rest/api/...` does not allow for server URLs of the format 
```
https://company.atlassian.net/wiki/
```
as the forward slash at the beginning  results here: https://github.com/AstroMatt/atlassian-python-api/blob/81b940d5859cf9dc311fa4e006ded0ea3d8aabe8/atlassian/rest_client.py#L42
to truncate the URL to `https://company.atlassian.net/` and loose the `/wiki`

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -31,16 +31,19 @@ class Confluence(AtlassianRestAPI):
         url = '/rest/api/content/{page_id}?expand={expand}'.format(page_id=page_id, expand=expand)
         return self.get(url)
 
-    def create_page(self, space, parent_id, title, body, type='page'):
+    def create_page(self, space, title, body, parent_id = None, type='page'):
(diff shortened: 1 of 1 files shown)


svyotov:
I just ran this:
```
status = confluence.create_page(
    space='TEST',
    parent_id=None,
    title='DEMO2',
    body='This is the body')
```
And this was the result:
image

AstroMatt:
Great, thank you very much for your contribution! :)
I have merged your code and I will push the new version to PYPI

AstroMatt:
New version 1.2.2 with your changes is already in PyPI.
Thanks!

PR #39, opened 2018-04-23
Fix - update_or_create function

opethe1st:
This function has this signature
```
def create_page(self, space, title, body, parent_id = None, type='page'):
```
so this call results in parent_id=space, page_id=title which is wrong. 
```
 result = self.update_page(parent_id, page_id, title, body)
```


AstroMatt:
That would be amazing!

PR #40, opened 2018-04-24
Log with log.error to get the errors from Atlassian

opethe1st:
screen shot 2018-04-24 at 20 05 52
 
It was a pain trying to figure out what the error was when I made a mistake. This should fix that. The screenshot is how an error will look now.

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -46,7 +46,6 @@ class AtlassianRestAPI(object):
             url += urlencode(params or {})
         if flags:
             url += ('&' if params else '') + '&'.join(flags or [])
-        # https://github.com/requests/requests/blob/e4fc3539b43416f9e9ba6837d73b1b7392d4b242/requests/models.py#L110-L122
         if files is None:
             data = json.dumps(data)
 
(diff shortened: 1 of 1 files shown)


AstroMatt:
I have one objection. You might find information in diff.

opethe1st:
I don't understand what you mean by "You might find information in diff". 
Right now, the request I made would just show up as 400 bad client with no additional information I can use to debug it.

AstroMatt:
error output should be used for errors.

```logging.getLogger('atlassian').setLevel(logging.INFO)```

this will help you change the level of verbosity of the logs if you want to debug.

**Comment by opethe1st** [comment]
True, but I just figured it would be better to get an error message that is useful by default when a ClientError occurs instead of just a 400 that doesn't mention the mistake I made .
I don't set up logging when I call libraries and I doubt most people do.

**Comment by AstroMatt** [comment]
Did you tested it with erroneous and positive responses?
As far as I understand it will spill error logs for successes which is very misleading.

**Comment by opethe1st** [comment]
Nothing gets printed when everything is alright 
screen shot 2018-05-10 at 21 15 12 . That is because - if everything is alright - the status code is a 200 or a 204 and so no error is logged. the relevant bits of code are here 
```
if response.status_code == 200:
            log.debug('Received: {0}'.format(response.json()))
        elif response.status_code == 204:
            log.debug('Received "204 No Content" response')
        else:
            self.log_curl_debug(method=method, path=path, headers=headers, data=data, level=logging.DEBUG)
            try:
                log.error(response.json())
            except json.decoder.JSONDecodeError:
                log.error(response)
            response.raise_for_status()
```
Or do you have a better idea how to make sure client errors get reported back? This is the best way I can think of, without any major changes to your code.

AstroMatt:
Now I get what you were trying to do. Thank you, and I am sorry it took so long.

PR #46, opened 2018-05-23
 Provide assignable users for project

gonchik:
Hi! 

I have added a new method. 
For fix use case, like generate the report who can be assignable on project. 
It is need for the security audit.

Cheers,
Gonchik Tsymzhitov


AstroMatt:
Thank you. New version (1.6.0) is already at pypi

PR #47, opened 2018-05-24
Add a new method

gonchik:
Hi! 

I have wrapped a group method. Just it is needed for audit users.
Small clean and add the parameter for the python 2.7.

Hope it will be needed. 

Cheers,
Gonchik Tsymzhitov


AstroMatt:
Thank you very much for your contribution!

PR #51, opened 2018-07-18
Confluence: Clean cache methods

gonchik:
Hi!

I have wrapped clean cache methods.
Cheers,
Gonchik Tsymzhitov


AstroMatt:
I like your pull request but I have two things:
- you've departed from API versioning which is not a good practice
- you've changed issuekey to issue_key. This is true with Python nature but... it is not consistent with Jira nomenclature. 

Of course if you have arguments I am always open to hear :)

gonchik:
Hi! 


> you've departed from API versioning which is not a good practice
I am going to update the method name. 

> you've changed issuekey to issue_key. This is true with Python nature but... it is not consistent with Jira nomenclature.
If we use issuekey instead of issue_key, in the future it will be so hard find the problems in rest api request. Because Jira nomeclature used also issueKey naming, also. 
In fact, PEP8 -> https://www.python.org/dev/peps/pep-0008/?#method-names-and-instance-variables
suggesting use the same rules method and instance variables. Hence now it is not consistens. 



Cheers,
Gonchik

PR #56, opened 2018-07-31
Bugfixes

gonchik:
Hi! 
This is bugfix release, related using params instead of in url. 
And raise IndexError if function will meet.

Cheers.
Gonchik Tsymzhitov


jforand:
This commit breaks the functionality of page_exists().

If you try to look up a page that doesn't exist using page_exists(), it delgates to get_page_by_title().

get_page_by_title() logs the error and returns None, but page_exists() doesn't perform a check on the return value, thus as a result of this commit, all pages exist

PR #67, opened 2018-08-21
Make more comfortable get_page_id behaviour

OrangeFlag:
If get_pade_id exec with wrong id parameter, don't raise 'AttributeError: 'NoneType' object has no attribute 'get''


gonchik:
Thanks!

PR #70, opened 2018-08-29
added CQL method to confluence

stranljip:
I saw the handy method for creating JQL queries in jira.py and decided to give it a try in order to have the same possibility for Confluence. It seems to work well:

`confluence.cql('space=TEST AND created > now("-10y") AND type=page',0,1)`


gonchik:
Thank you for PR

PR #79, opened 2018-09-24
Method to get the page_id if we have the title of the pge, parent page id and space name

AdityaMisra:
Method to get the page_id if we have the title of the pge, parent page id and space name

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -563,3 +563,24 @@ class Confluence(AtlassianRestAPI):
         if cql is not None:
             params['cql'] = cql
         return self.get('rest/api/search', params=params)
+
+    def get_descendant_page_id(self, space, parent_id, title):
+        """
+        Provide  space, parent_id and title of the descendant page, it will return the descendant page_id
+        :param space: str
+        :param parent_id: int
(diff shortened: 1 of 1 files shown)


atlassian/confluence.py
```diff
@@ -563,3 +563,24 @@ def cql(self, cql, start=0, limit=None):
...
+        """
+        Provide  space, parent_id and title of the descendant page, it will return the descendant page_id
+        :param space: str
+        :param parent_id: int
+        :param title: str
+        :return: page_id of the page whose title is passed in argument
+        """
+        page_id = ""
+
+        url = 'rest/api/content/search?cql=parent={}%20AND%20space="{}"'.format(
```
gonchik: Hi! 

Thank you for your PR. 
Have you checked related info include or not, in returned json from get content method?

AdityaMisra: Hey
In the returned json we have 'results', 'start', 'limit', 'size', & '_links' keys. 
In results we have list of descendant pages.
Is this what you were asking me?

gonchik: Yeap!

PR #84, opened 2018-10-04
So old PR related Confluence export as pdf  example

gonchik:
Hi! 

This PR include example method for Confluence

Cheers,
Gonchik Tsymzhitov


gonchik:
This PR include some changes for return method.

PR #87, opened 2018-10-09
Added: add_attachment function

SLRover:
Updated: docs, add add_attachment section


SLRover:
Cancel. An error was found.

PR #91, opened 2018-10-18
add function to create label on page for confluence

fpatou:


gonchik:
Thanks for PR. 
We will wait a new things

Issue #101, opened 2018-10-26
The paged requests are not managed


danielesiddi:
Atlassian API returns a paged JSON answer.
This shoulde be managed by wrapper too, extending array with all the results.

SLRover:
Hello!
What method are you using?

SLRover:
I checked it. The request **get_tags** returns a list of dictionaries. Not JSON.
What type of answer are you want to get from API wrapper?

danielesiddi:
Because the atlassian API returns a json with maximum 1000 elements.
If you have more than 1k tag (which, in big projects, is probable), you just receive the first 1k.
So, I think the library should check the isLastPage flag in json API, and join all the values together.

Please check the example Bitbucket JSON
`{"size":25,"limit":25,"isLastPage":false,"values":[]}`

The maximum admitted for limit is 1000

SLRover:
OK! Thanks for the improvement proposal.

gonchik:
@danielesiddi  Please, have a look PR #102 

Now it should be works well.
I have added in log.info level related messages 
Also, use condition like this len(stash.get_tags()) > 0 

Cheers,
Gonchik Tsymzhitov

gonchik:
@danielesiddi  Any tests from your side?

danielesiddi:
Hi. Going to do it today.

gonchik:
@danielesiddi  
I have tested. 
You can see the example here 
https://github.com/gonchik/atlassian-python-api/commit/8a1563ca2b93885869f25a5ed547f703a32913fb

PR #102, opened 2018-10-26
Set the start parameter and the log info messages

gonchik:
Hi! 

this is bugfix for get_tags method in bitbucket module.
Also, I have added a log messages 

Cheers,
Gonchik Tsymzhitov


gonchik:
#101  will be fixed

PR #122, opened 2018-11-07
Added: New methods for Jira module (managing users in groups)

SLRover:
Updated:
 - Docs (added description for new methods)
 - REST Client (added 'params' argument into POST method)


gonchik:
Thanks for the PR

PR #146, opened 2019-01-21
Jira.jql() api changes

AstroMatt:
* As of limit is by default 50 set in Jira, I removed None value and used 50 as a default argument
* Allowed fields to be passed as a list, in that case it will convert it into comma separated values


gonchik:
Hi! 

Thanks for your PR. 
Can we stay as limit parameters as None? Because if you request without that, will set automatically default value, And it can be adjusted by administrator in configuration. 

About comma separated, that's so good. 

Thanks 

Cheers,
Gonchik Tsymzhitov

PR #162, opened 2019-03-12
Bitbucket: added enable, disable, get, set methods for branching model

approximatenumber:
Added new methods for branching model, docs updated (example in [docs](https://github.com/atlassian-api/atlassian-python-api/blob/bef94b43194dfc56d1530aa9861d221aaef7f2fa/docs/bitbucket.rst#manage-repositories)).


gonchik:
Thanks for you PR!

PR #166, opened 2019-03-19
Issue: NameError: name 'log' is not defined

diogowatson:
Error: NameError: name 'log' is not defined

cause: logging referenced as log in jira.py but not imported as it.
solution:
import logging as log

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -1,5 +1,5 @@
 # coding: utf8
-import logging
+import logging as log 
 from requests.exceptions import HTTPError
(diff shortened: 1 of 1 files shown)


gonchik:
Hi! 
Why do you think it is related to that bug? 
as I see the next changes will generate errors
```
import logging as log
log = logging.getLogger(__name__)
```

gonchik:
I hope that PR will fix your request #169

gonchik:
@diogowatson  Is it covered already?

gonchik:
During inactivity I have closed that PR

PR #168, opened 2019-03-28
Add session parameter to rest_client

seanson:
This allows users to pass a custom requests.Session to the rest_client.
This enables things like pre-configuring a session for SSO auth for
servers that do not allow basic HTTP auth.


gonchik:
Thanks for that!

PR #175, opened 2019-04-10
Ability to add a comment

Woklex:
New method add_comment. There is an option to add a comment to Jira, why not make the option to add a comment to the wiki?


gonchik:
Thanks! Really, why not :)

PR #176, opened 2019-04-11
Get labels

Woklex:
Perhaps, not only it will be useful for me to receive labels.
Of course, you can get them like this:
Jira.issue("Jira-5116")['fields']['labels']
But it is more convenient:
Jira.get_issue_labels("Jira-5116")


gonchik:
Hi! 
Got it! :) 

Thanks for that

Cheers,
Gonchik Tsymzhitov

PR #191, opened 2019-04-23
Jira: user_find_by_user_string(). Add "include_active_users" param.

djachkov:
WIth include_inactive_users you get both active and inactive users. Now you can make a request only for inactive users.


gonchik:
Hi! 
Thanks for that point. 
As I see in Enterprise release 7.6.x you met, fortunately on 7.3.x it worked. 
That's make sense. 

Cheers,
Gonchik Tsymzhitov

PR #192, opened 2019-04-23
Added delete_issue_remotelink_by_link_id

austinmh12:
Updated the docs to match as well. 

Unfortunately, I haven't found a way to prevent the printing of the response in create_or_update_remotelinks. I don't know if there's a way to suppress that that I'm just not seeing, or if it's just the way it works, so I hadn't updated it. If you know how to, feel free to add that or let me know how to do it.

I plan on adding the application portion to the create_or_update_issue_remotelinks, but I can only access Jira and Confluence while at work and our Jira is in the process of being upgraded so the Jira to Confluence link is broken at the moment, so I'm unable to test.

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -621,6 +621,11 @@ class Jira(AtlassianRestAPI):
         return self.post(url, data=data)
     
     def delete_issue_remotelink_by_link_id(self, issue_key, link_id):
+        """
+        Deletes Remote Link on Issue
+        :param issue_key: str
+        :param link_id: str
+        """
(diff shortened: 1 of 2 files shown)


gonchik:
Hi! 

Typically I prevent by logging functionality e.g.
logging.basicConfig(level=logging.ERROR)

I have renamed method name from create_or_update_issue_remotelinks to create_or_update_issue_remote_links . Because in that more clear and foundable by IDE, e.g. PyCharm

I will mention you in that PR 

Cheers,
Gonchik Tsymzhitov

PR #193, opened 2019-04-23
Polish naming of method and bump version

gonchik:
Hi! 
That PR include polish naming of method. (Sorry, for that, better way is rename as soon as possible instead of later)

Cheers,
Gonchik Tsymzhitov


gonchik:
@austinmh12  Please, be informed

PR #203, opened 2019-05-08
Implemented new bitbucket project method

pablotoledo:
The implemented method allows to grant project permissions to an specified user


gonchik:
Thanks for that PR

PR #205, opened 2019-05-10
Add Confluence markup format parameter

yoanngoular:
It will allow to create Confluence pages that will be interpreted
with the wiki markup syntax.

More info about Confluence formats here:
https://confluence.atlassian.com/doc/confluence-markup-283640216.html

Change-Id: I2534634a1b1eec1b9d970e5a055b8e73735fd8c9


gonchik:
Thanks for that PR :)

PR #206, opened 2019-05-10
Delegate http responses handling to the caller

yoanngoular:
The interpretation of the REST API responses is not
part of the responsibility of this library.
Thus, it has been delegated to the user.


yoanngoular:
Returning the response as a JSON to the caller was a real problem to interpret it on the user's side. The full Response object is perfect so we can use the response.status_code and all the other attributes it offers to react accordingly. Finally, the logs displayed depending on the status code are clearly something the user of the library should take care of and not the library itself. I hope you'll find this change useful. On my side, I would love to have it merged :smile:

gonchik:
Hi @yoanngoular ! 
Thanks for your PR, can we a little bit polish it?:)
Could you show me your use cases? 
Can we wrap into a new class rest client py with normal exceptions etc ? 
Because sometimes good to have advanced mode, power mode and user mode for logging and response manage. 

I would like to review exist one, wrap into different modes.

Cheers,
Gonchik Tsymzhitov

yoanngoular:
Hi @gonchik,
Of course, we can modify it if you want to.
The use case is simple, let me ask you this: 
Let's say you are developing a python app that uses your library to perform Confluence requests.
How do you manage the different HTTP responses in your application using your library?
You can't... Because it's already abstracted by the library so you can't manage it the way you want, you are currently forced to have some stuff logged by your library and you can't catch the HTTP response code since you just get the response as a JSON back.

That's why I want to delegate that responsibility to the user of the library so he is free to do whatever he wants with the different HTTP status codes and responses.

Do not hesitate to tell me if it's not clear to you.

Regards,
Yoann

yoanngoular:
Up :)

gonchik:
@yoanngoular  what do you think? 
if  evaluate for the AtlassianRestAPI class the field `self.request_response`. 
Which can be populated by `self.request_response = response`, for every request? 

When advanced user can reuse that response to check the last response ?
like `status_code = confluence.request_response.status_code`

yoanngoular:
I don't think that this is a good idea. Why don't you want to let access to the response object? I mean, there is no regression here, people that only want the JSON body of the response would just have to call response.json(). And every one that needs the full object (I think that in most case you do need that full object) can have access to it directly because it would be returned by the function as is.

gonchik:
Just a few thing
From my side a lot sanity checks in our CI systems based on exist errors in logs. 
or let's evaluate a field advanced mode, where we will return the exist object  ?

yoanngoular:
I did not fully understand your answer. About the "advanced mode argument", why not but I still don't think it's the right solution. It's a responsibility issue here in my humble opinion.
Can you do the modification on your side then?

Regards

gonchik:
Hi @yoanngoular 
What do you think this way?
https://github.com/atlassian-api/atlassian-python-api/commit/f845183d7cd55b1f1e6599e32ec9fe62efa6f1aa
sorry for the delay

PR #210, opened 2019-05-14
Added 4 new methods in Jira

austinmh12:
Added create_filter for creating jira filters.

Added 3 methods for agile:
    create_agile_board to create boards
    get_agile_board_by_filter_id as another way to search for a board
    get_issues_for_backlog to return a paginated list of backlogged issues on a board


gonchik:
@austinmh12  Thanks!

PR #220, opened 2019-05-21
Add set_issue_status_by_id

Woklex:
It is better to close the task by ID, the status name can be changed :(


gonchik:
Thanks! :)

PR #223, opened 2019-05-23
Update jira.py

HoldemGK:
Hi. Added several methods that were needed for my project.


gonchik:
Thanks for that PR. 
Could you please, add a few polish things ? 
Also, 'Value' -> may be 'value' ?

HoldemGK:
> Thanks for that PR.
> Could you please, add a few polish things ?
> Also, 'Value' -> may be 'value' ?

I am currently working on migration to Jira. This is only the first stage. I will be glad to add something as needed in the future.

PR #231, opened 2019-06-05
Adding ability to pass cookies

mduaneh:
In my Confluence instance I needed to be able to inject different cookies.  This change allows Cookies to be declared during creation of the AtlassianRestAPI object.

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -17,7 +17,7 @@ class AtlassianRestAPI(object):
                           'X-Atlassian-Token': 'no-check'}
 
     def __init__(self, url, username=None, password=None, timeout=60, api_root='rest/api', api_version='latest',
-                 verify_ssl=True, session=None, oauth=None):
+                 verify_ssl=True, session=None, oauth=None, cookies=None):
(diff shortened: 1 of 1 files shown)


gonchik:
Hi! 
@mduaneh  Have you checked on your side that PR, please?

gonchik:
@mduaneh  What do you think to set the method to extend the session parameters by set the headers?

mduaneh:
@gonchik Sorry I do not fully understand your question?
If you mean could this have been accomplished with headers, I couldn't figure out how.
I would have to respecify all the default headers as well, making the one addition more difficult/cumbersome.  Also, the rest_client's use of curl was only expecting cookies to be specified with -b indicating it is a completely separate entity.  (be aware, I'm no http expert)

https://en.wikipedia.org/wiki/HTTP_cookie

mduaneh:
Also, here is an example of how I used it with my modifications:
cookies={"name":"value"}
confluence = Confluence(
    url='https://myconfluence.site.com',
    cookies=cookies
    )

gonchik:
Got it

PR #235, opened 2019-06-07
add representation support to all confluence actions modifying content

wrenoud:
create_page already support this, so support is extended to:
* update_page
* update_or_create
* update_existing_page
* append_page


gonchik:
Thanks

PR #245, opened 2019-06-18
Fix incorrect variable name

djachkov:
Yes, I needed to do 4 commits to fix the incorrect variable name. =(


gonchik:
I would say the result the same, just polish! Thanks for that

PR #265, opened 2019-07-11
add one new method to get issue all fields

mmh891113:
Signed-off-by: Mamh-Linux <bright.ma@blackshark.com>


mmh891113:
for this method we can send one http request and get all fields we want. then get each field like project,summary,issuetype....

PR #274, opened 2019-07-16
Fix undefined name 'permission' in project_remove_user_permissions

aK0nshin:
Can't find any 'permission' parameter to implement in documentation.
https://docs.atlassian.com/bitbucket-server/rest/4.5.1/bitbucket-rest.html#idp2397808


gonchik:
Thank you!

PR #282, opened 2019-07-18
Issue 279: bug fix for upload a new version of an attachment 

andreas-j-hauser:
Here is my bugfix for the reported bug. Now it is possible to upload a new version of the file. The name of the attachment is only the file name and doesn't include the complete path.

I added a test case to reproduce the bug (without the fix) and verify the fix.


gonchik:
Thank you for unittests  as well :)

gonchik:
After 10 min version will be upload

PR #288, opened 2019-07-23
Issue 286: Confluence.get_page_by_title() works like Confluence.get_page_by_id()

andreas-j-hauser:
Attention: If anyone use parameter start and limit to call Confluence.get_page_by_title() he will get an TypeError exception

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -103,9 +103,10 @@ class Confluence(AtlassianRestAPI):
                  If it has IndexError then return the None.
         """
         url = 'rest/api/content'
-        params = {}
-        params['start'] = 0
-        params['limit'] = 1
+        params = {
+            'start': '0',
+            'limit': '1'
+        }
         if expand is not None:
             params['expand'] = expand
         if space is not None:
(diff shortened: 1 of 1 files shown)


atlassian/confluence.py
```diff
@@ -92,24 +92,23 @@ def get_page_space(self, page_id):
         """
         return ((self.get_page_by_id(page_id, expand='space') or {}).get('space') or {}).get('key')
 
-    def get_page_by_title(self, space, title, start=None, limit=None):
```
gonchik: @andreas-j-hauser  Can we just set into method variables def get_page_by_title(self, space, title, expand=None, start=0, limit=1) ?

andreas-j-hauser: Yes we can. I see the benefit of your suggestion.

But I'm not so happy with this solution, because using of the parameters brings trouble.
Do you think it helps do not remove parameter start and limit, but ignore the values?
It is possible to mark the parameter as obsolete?

gonchik: About ignore, it will work ok. 
about obsolete, I see only decorator, but it is information as is.
https://stackoverflow.com/questions/2536307/decorators-in-the-python-standard-lib-deprecated-specifically

PR #291, opened 2019-07-25
Add get_issue_transitions_full method to Jira class

aK0nshin:
Docs:
https://docs.atlassian.com/DAC/rest/jira/6.1.html#d2e1078

I don't understand why you add 'fields' by default if you don't use them in return.


gonchik:
Hi! 
I would say I have implement on my automatisation. 
And after that it stopped. 
Thanks for all your improvements

Cheers,
Gonchik Tsymzhitov

aK0nshin:
You are welcome =)

PR #298, opened 2019-07-31
Confluence: Add expand parameter to get_all_spaces

AlexanderLill:
The 'expand' parameter is defined for the API, but until now the code did not support it for this particular function.


gonchik:
Thanks!

PR #308, opened 2019-08-08
Fix #307 (get_pages_by_label)

### LINKED ISSUE #307: HTTP 400 when using get_all_pages_by_label
When trying to using get_all_pages_by_label I get an HTTP 400 error returned from confluence. It looks like this is because the parameters are being url encoded twice:

`DEBUG:urllib3.connectionpool:https://confluence:8090 "GET /confluence/rest/api/content/search?cql=type%3Dpage%2520AND%2520label%3Dtestlabel&limit=50`

The parameters should be:

`cql=type%3Dpage%20AND%20label%3Dtestlabel`

(i.e. the % in %20 is being encoded as %25)

jdsnape:
Fix #307

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -216,7 +216,7 @@ class Confluence(AtlassianRestAPI):
         url = 'rest/api/content/search'
         params = {}
         if label:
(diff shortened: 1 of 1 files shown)


gonchik:
Thank you for that caught!

PR #312, opened 2019-08-15
add limit restriction for get_all_pages_from_space

dacog:
I tested and found that the max number of articles one can get by calling *get_all_pages_from_space* is 100. I think it would be useful to add that to the documentation.

### DIFF
--- a/docs/confluence.rst
+++ b/docs/confluence.rst
@@ -35,8 +35,9 @@ Get page info
 
     # Get all pages from Space
     # contet_type can be 'page' or 'blogpost'. Defaults to 'page'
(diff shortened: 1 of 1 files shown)


gonchik:
Good job!

DivyaWadehra:
How  to increase the limit? It's a static number

gonchik:
@DivyaWadehra I do recommend you to make iteration for get all pages

tpatel-leena:
@gonchik 

iteration can only be made when we know that there is a next page available.
we are just sending results as a response, no other information is being send for now
I have raised a PR for it
https://github.com/atlassian-api/atlassian-python-api/pull/1090

PR #330, opened 2019-09-13
Docs: Correct documentation of Confluence.update_page()

magnusbaeck:
Previously the documentation erroneously indicated that Confluence.update_page() required the parent page id as the first positional argument when it actually is a kwarg. Also add the missing minor_edit kwarg.


gonchik:
Thanks!

PR #345, opened 2019-09-25
New example added 

copolycube:
Added a new example for JIRA that :
* loops over each projects
* creates a component datastructure with the minimal informations required
* create the components in each projects (same list of components)


gonchik:
Thanks

PR #351, opened 2019-09-27
Check if the size exists before actually indexing it DLS

sed-sytarchuk:
I came about this while trying to use the api to upload attachments to a page that didn't have any attachments associated to it.

I got a KeyError on this line and changing it to use .get instead of indexing it fixed it for me.


gonchik:
Thanks

PR #352, opened 2019-09-30
Log level of package imports adjusted

offa:
The error message on the first import is a little misleading on windows: it appears always, but everything is ok. I'm not sure if is is needed anyway since the import error is a valid case on windows.

For the second message: Does it make sense to throw an error here? At this point it's not possible to continue with a valid state.


gonchik:
Thanks, make sense

PR #354, opened 2019-10-01
Kerberos follow-up

offa:
Some open issues mentioned in #353 fixed:

1. Exception if kerberos is used but not installed
1. Tox config updated to include the extra dependency
1. Note added to the documentation. Is this enough or is a full pip call more suitable?

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -56,8 +56,7 @@ class AtlassianRestAPI(object):
             try:
                 import kerberos_sspi as kerb
             except ImportError:
-                log.error("Please, fix issue with dependency of kerberos")
-                return
+                raise ImportError("No kerberos implementation available")
         __, krb_context = kerb.authGSSClientInit(kerberos_service)
         kerb.authGSSClientStep(krb_context, "")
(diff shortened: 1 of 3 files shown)


offa:
The installation of optional dependencies through the pypi package doesn't work.

```sh
> pip install atlassian-python-api[kerberos]
```
Results in _No matches found_.

It works from source and the installed package contains some sort of dependency information (`requires.txt`) :thinking: :

```ini
requests
six
oauthlib
requests_oauthlib

[kerberos]

[kerberos:platform_system != "Windows"]
kerberos

[kerberos:platform_system == "Windows"]
kerberos-sspi
```

offa:
Well … quotes to the trick:

```sh
pip install "atlassian-python-api[kerberos]"
```

And it works! :+1:

offa:
There's just on issue left: How to handle the `requirements.txt`?

1. Move kerberos to a seperate one which sources the common file
1. Keep all dependencies in one file as it's at the moment

Are there any preferences?

gonchik:
I think the better way is keep in one file, as we can number of dependency.

offa:
So, then the PR is ready from my side.

gonchik:
Thank you :)

PR #374, opened 2019-10-25
added get/set repo settings

lelandsindttouchnet:
Added get and set repo settings functions. 

I am considering more specific functions, but figured I would start here and get feedback.

### DIFF
--- a/atlassian/bitbucket.py
+++ b/atlassian/bitbucket.py
@@ -553,6 +553,27 @@ class Bitbucket(AtlassianRestAPI):
         data = {"name": str(name), "endPoint": str(end_point)}
         return self.delete(url, data=data)
 
+    def get_pull_request_settings(self, project, repository):
+        """
+        Get pull request settings.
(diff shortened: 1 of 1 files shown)


gonchik:
Hi! That's good for start :)

lelandsindttouchnet:
Well, ok. ;)

Which to do you think would provide more value.... 

building out `set_pull_request_minimum_approvers` or providing an example using `set_pull_request_settings`

```
from atlassian import Bitbucket
bitbucket = Bitbucket(
        url='http://localhost:7990',
        username='admin',
        password='admin')
data={}
data['requiredApprovers'] = 1
response = bitbucket.set_pull_request_settings(project='AP', repository='approve', data=data)
```

gonchik:
So it's depends on. 
BTW, you can find somewhere I already wrap a few things like you. 
sometimes, it's good use list arguments
https://realpython.com/python-kwargs-and-args/

PR #375, opened 2019-10-28
bug in variable not None check

svyotov:


gonchik:
Thanks! 
Yeah, it's mistake

PR #379, opened 2019-10-31
fix/change the parameter filter of get_tags to be filterText

krysalead:
Hi,

according to this page
https://metacpan.org/pod/WebService::BitbucketServer::Core::V1#get_tags
the name of the parameter for the filtering is filterText.


gonchik:
Thanks

PR #382, opened 2019-11-02
new method added to stop build which is in progress

dunterov:
Equivalent to the "Stop Build" button in GUI


gonchik:
Hi! 
Thanks, I hope that's method don't need the change exist headers :)
Cheers,
Gonchik Tsymzhitov

PR #385, opened 2019-11-13
Fix format error in log 400 code error

MyMaloks:
Whenever the HTTP response is with error code 400, some info is logged.
But there are two variables for only one {}. That makes the format function fail.


gonchik:
Thanks :)

PR #387, opened 2019-11-14
Add 2 application related functions for Jira

MyMaloks:
The REST API `rest/api/2/user/application` was not implemented.
The PR aims at adding it.

It also add an utility function to easily determine whether the user already has a specified application role.


gonchik:
Thanks!

PR #410, opened 2019-12-06
Confluence: Referencing none-existing pages will return None

masics:
Confluence: Referencing none-existing pages will return None instead of throwing a runtime error


gonchik:
Thank you!

PR #413, opened 2019-12-17
Wrap label in quotes to handle labels with dashes.

dmort-ca:
Using get_all_pages_by_label returned a 400 when the label included dashes (e.g. test-label-with-dashes).

Wrapped the label in quotes when building the cql query.


gonchik:
Thanks!

Issue #317, opened 2019-08-23
Cannot update only title of the page


Vanderhoof:
If I use `Confluence.update_page` but only want to change title, so I supply the old body and a new title, I get:

```
Content of <page_id> is exactly the same
```

Probably it should check for changes not only in body but also in the title of the page.

gonchik:
It has been fixed

PR #414, opened 2019-12-19
Include title in determining if there is a difference

jayudey-vertex:
## Problem
I hit this issue, https://github.com/atlassian-api/atlassian-python-api/issues/317 as well when I was trying to update just the title of a confluence page. The current method in determining if an update only considers the body content of the confluence page.

## Solution
Include the title in comparing page content.

## Areas of impact
Updating confluence pages.


gonchik:
It sounds good :)

PR #424, opened 2020-01-08
Add append option to the begining of the page

Ryukamusa:
I don't know if this change makes sense but I was using your api and a method was useful to add stuff to the beginning of an existing page similar to append. Therefore I made it. I considered having it as a flag on append method and make it possible to append on top of page. I can do it like that if it makes more sense

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -785,6 +785,45 @@ class Confluence(AtlassianRestAPI):
 
             return self.put('rest/api/content/{0}'.format(page_id), data=data)
 
+    def insert_to_existing_page(self, page_id, title, insert_body, parent_id=None, type='page', representation='storage',
+                    minor_edit=False):
+        """
+        Insert body to the begining of page if already exist
+        :param parent_id:
+        :param page_id:
+        :param title:
+        :param insert_body:
+        :param type:
+        :param representation: OPTIONAL: either Confluence 'storage' or 'wiki' markup format
+        :param minor_edit: Indicates whether to notify watchers about changes.
+            If False then notifications will be sent.
+        :return:
+        """
+        log.info('Updating {type} "{title}"'.format(title=title, type=type))
+
+        if self.is_page_content_is_already_updated(page_id, append_body, title):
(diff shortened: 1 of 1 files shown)


gonchik:
First of all, thank you for your PR.
Can we add to method, like append_to_page, prepend_to_page and last one method is, insert to page with flag, like your
https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/confluence.py#L788

Thanks, 
Gonchik

Ryukamusa:
Thanks for the reply.
I am sorry but I am not sure I understood what you meant. So my initial idea was either having the method that could be renamed to prepend_to_page as it is but that will create some duplicated code (like it did). 
Therefore the second option was changing the method append_page to `def append_page(self, page_id, title, append_body, parent_id=None, type='page', representation='storage', minor_edit=False, top_of_page=False):
....
body = append_body  + previous_body if top_of_page else previous_body + append_body
`.
The rest of the method would be exactly the same and the duplicated code would not be necessary.

gonchik:
So I have the same idea :) 
Just in **JS**, we can use append and prepend  method names. 
Both method will reuse the name method: 
`def insert_to_existing_page(self, page_id, title, append_body, parent_id=None, type='page', representation='storage', minor_edit=False, top_of_page=False)`

as final we will have like this code
`def append_page(self, page_id, title, append_body, parent_id=None, type='page', representation='storage', minor_edit=False)
     return insert_to_existing_page(self, page_id, title, append_body, parent_id=None, type='page', representation='storage', minor_edit=False, top_of_page=False)`
and 
`def prepend_page(self, page_id, title, append_body, parent_id=None, type='page', representation='storage', minor_edit=False)
     return insert_to_existing_page(self, page_id, title, append_body, parent_id=None, type='page', representation='storage', minor_edit=False, top_of_page=True)`

Ryukamusa:
So I changed it to have two methods append and prepend and both use the same _insert_to_existing_page method which has the top_of_page keyword argument

gonchik:
Thank you!

PR #425, opened 2020-01-13
Fix few get requests in ServiceDesk API falls if advanced_mode is on

MaximZemskov:
Hi. Few methods in `ServiceDesk` api falls if `advanced_mode` is `True`, because they expecting data and not `Request` obj.


gonchik:
Thank you! 
a new release will be uploaded soon

Issue #232, opened 2019-06-05
Bitbucket API v2


oscarverona:
Hello,

It seems that Bitbucket removed the version 1 of their API on May 29. I cannot make it work. ¿Is it planned to update the api client to version 2?

Thanks

gonchik:
Hi! 

Sure!

SilverJan:
Hi, by when can we expect the update?

tuxmaster5000:
It looks like to get complicated, because V2 are only supported on the cloud version of it.
The self hosted one use V1. Like it is called here:
https://community.atlassian.com/t5/Bitbucket-questions/Does-the-rest-api-2-0-exist-for-server/qaq-p/881631
Or have someone other information's?

gonchik:
@tuxmaster5000  Thanks ! 
@SilverJan  @oscarverona I hope you will raise a PR for the improvements :)

PR #428, opened 2020-01-20
Prepare for Bitbucket API 2.0

tuxmaster5000:
My idea for https://github.com/atlassian-api/atlassian-python-api/issues/232
It will only contains the base part, so it will not break anything.  ( I hope)
The API 2.0 endpoints must be added, because I can't test it without an cloud account.

### DIFF
--- a/atlassian/bitbucket.py
+++ b/atlassian/bitbucket.py
@@ -30,7 +30,7 @@ class Bitbucket(AtlassianRestAPI):
         """
         if self.standalone:
             url = 'rest/api/1.0/projects/{0}'.format(key)
-        else
+        else:
              url = 'rest/api/2.0/projects/{0}'.format(key)
         return self.get(url) or {}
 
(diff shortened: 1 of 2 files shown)


gonchik:
Hi!
Thanks for your idea  and preparation. 
What do you think if we will use  a field cloud instead of  standalone in rest class?
Of course it's easy win :) to be compatible with cloud API. 

Also, what do you think reuse that lib as git submodule and small overwrap of auth. 
https://bitbucket.org/atlassian/python-bitbucket/src/master/pybitbucket/auth.py
?

or we can continue use the 2 versions of end-points

tuxmaster5000:
I have take an look at:
https://bitbucket.org/atlassian/python-bitbucket/src/master/pybitbucket/auth.py
I think here using an wrapper will let us ran in trouble, because not all endpoints in the source starts with: rest/api/VERSION ... I have seen some calls that shows to rest/insights, rest/git-lfs/ and so on.

mashayev:
Any plan to merge it soon?

gonchik:
@tuxmaster5000  Thanks for the PR. 
I have a little bit inversed the field of rest client api as mostly lib works with standalone.

PR #433, opened 2020-01-24
Confluence: fix #412

### LINKED ISSUE #412: [Confluence] AttributeError: 'Response' object has no attribute 'get' in update_page in advanced mode
Hi,
I get an error while calling `Confluence.update_page`  in advanced mode:
```
  File "/.../venv/lib/python3.6/site-packages/atlassian/confluence.py", line 705, in update_page
    if self.is_page_content_is_already_updated(page_id, body):
  File "/.../venv/lib/python3.6/site-packages/atlassian/confluence.py", line 665, in is_page_content_is_already_updated
    confluence_content = (((self.get_page_by_id(page_id, expand='body.storage') or {})
AttributeError: 'Response' object has no attribute 'get'
```
looks like  code at [this line](https://github.com/atlassian-api/atlassian-python-api/blob/88156a507873eea75d52436f09ff872435590450/atlassian/confluence.py#L696) expects that `self.get_page_by_id` always returns `dict` object but  [AtlassianRestAPI.get](https://github.com/atlassian-api/atlassian-python-api/blob/88156a507873eea75d52436f09ff872435590450/atlassian/rest_client.py#L191) method, which is used inside of `get_page_by_id`, can also return `Response` in advanced mode or `str` in case of any errors.

djachkov:
This is a quick fix for an issue. I think we should make a review of the advanced_mode features for all methods. 
It not the first time, when some troubles appear because of that.

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -712,9 +712,14 @@ class Confluence(AtlassianRestAPI):
         :param title: Title to compare
         :return: True if the same
         """
-        confluence_content = (((self.get_page_by_id(page_id, expand='body.storage') or {})
-                               .get('body') or {})
-                              .get('storage') or {})
+        if self.advanced_mode:
+            confluence_content = (((self.get_page_by_id(page_id, expand='body.storage').json() or {})
+                    .get('body') or {})
+                    .get('storage') or {})
+        else:
+            confluence_content = (((self.get_page_by_id(page_id, expand='body.storage') or {})
+                    .get('body') or {})
+                    .get('storage') or {})
 
         if title:
             current_title = confluence_content.get('title', None)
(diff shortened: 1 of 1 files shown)


atlassian/confluence.py
```diff
@@ -712,9 +712,14 @@ def is_page_content_is_already_updated(self, page_id, body, title=None):
         :param title: Title to compare
         :return: True if the same
         """
-        confluence_content = (((self.get_page_by_id(page_id, expand='body.storage') or {})
-                               .get('body') or {})
-                              .get('storage') or {})
+        if self.advanced_mode:
+            confluence_content = (((self.get_page_by_id(page_id, expand='body.storage').json() or {})
```
mikitachab: Hi!
maybe better will be something like that:
```py
page = self.get_page_by_id(page_id, expand='body.storage') 
page = page.json() if self.advanced_mode else page
confluence_content = (((page or {})
              .get('body') or {})
              .get('storage') or {})
```
it helps to avoid code duplication and makes code more readable.

atlassian/confluence.py
```diff
@@ -765,7 +770,10 @@ def update_page(self, page_id, title, body, parent_id=None, type='page', represe
             return self.get_page_by_id(page_id)
         else:
             try:
-                version = self.history(page_id)['lastUpdated']['number'] + 1
+                if self.advanced_mode:
```
mikitachab: and the same here
```py
page_history = self.history(page_id).json() if self.advanced_mode else self.history(page_id)
version = page_history['lastUpdated']['number'] + 1
```

gonchik:
@mikitachab Would you like to provide a some PR for refactoring ? :)

mikitachab:
@gonchik no problem, I will find some time and do it later.

PR #442, opened 2020-02-07
Added default reviewers methods for Bitbucket API

adribaeza:
Create standar methods get, create, update and delete conditions (reviewers) for projects and repositories for Bitbucket API


gonchik:
@adribaeza  Thank you! 
 I will be happy if you provide a methods for cloud as well :)

Issue #257, opened 2019-07-01
Implement better error handling


bradphelan:
For some reason the JIRA server is randomly throwing authentication errors. I'll take this up with my IT group. however the python API doesn't throw any exceptions it just returns a string of the html of the error. For example if I do
```
jira = Jira(
    url=host,
    username=user,
    password=passwd)

data = jira.jql(q)
```

if there is an auth error then data comes back as the string content of the html error from jira. It would be nice if the python api would just detect this internally and throw some kind of exception. Otherwise there is a cryptic error later on where I try to index the data structure and get a failure because it is not the correct type.

gonchik:
Hi! 
Thanks for that, I am thinking to redesign exist handlers. 
It will be good if you provide more detail info, like logs etc. 
and your expectation as well

tntC4stl3:
Agree, I think it's better to let the caller to decide how to handle those exceptions instead of only log in api library.

tntC4stl3:
In my case, sometimes the IT changes the field attribute without notifying users, so it logs error in `atlassian-python-api`, but the caller doesn't know this is already failed.
![Screen Shot 2019-07-22 at 4 46 35 PM](https://user-images.githubusercontent.com/1506238/61619043-a9801880-aca0-11e9-864f-e5df60b5200f.png)

Vanderhoof:
Also I would suggest to make responses more consistent.
For example:

* If I call `Confluence.get_page_by_id` and provide the wrong ID, i get a response dictionary with `statusCode` and `data` fields.
* If I call the same method with the correct ID — I only get the page dictionary, without any `statusCode` fields. So to check for errors I have to check if the `statusCode` field is in the dictionary first?
* Next, if I call the same method, but it turns out that the login or password provided were incorrect. In this case I will always get a string with error page HTML. So it means that first I have to check that the type of the response, and then look for `statusCode`? That seems a bit wrong.

Maybe there are some other reponse options, I just started using the lib.

I think (I may not know everything) that it would be best to always have similar response dictionary structure with `statusCode` always present for easy error checking. I imagine that changing the working logic (in my example when the id was provided correctly) would be painful for users, because they've already have a lot of code using the old structure. But at least it should be possible to make all error responses consistent. Especially because the string with error page I got in the last scenario had an HTTP status code and error message written inside.

tbugfinder:
I'd like to second that this python-lib should raise exceptions if the server returns an error.

[...]

PR #447, opened 2020-02-11
Improve error handling

rlmaers:
This is a suggestion for implementing better error handling as noted in #257.

**Overview**
* rest_client.AtlassianRestAPI.request: Propagate errors from requests by using the `raise_for_status` method in the response object.
* errors: Define custom exception classes.
* confluence: Use exception chaining to raise the custom exceptions to convey more useful information when possible.

**Notes**
* The improved error handling is implemented for Confluence only right now. This is the Atlassian product I use (and can thus test) most right now. It also appears to have a smaller API surface than Jira for instance.
* I wasn't completely sure about the coding style, but attempted to shorten the exception messages to a reasonable length.
* The names of the custom exception classes could probably be better, but this is probably easier to improve once more products and cases are implemented.
* The error messages are (almost) verbatim copies from the API documentation, but could perhaps also be normalized.

I don't expect this to be merged as is. I only made the changes and this pull request to test and show this approach.

Any feedback appreciated.

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -6,6 +6,13 @@ import time
 from requests import HTTPError
 
 from atlassian import utils
+from .errors import (
+    ApiError,
+    ApiNotFoundError,
+    ApiPermissionError,
+    ApiValueError,
+    ApiConflictError
+)
 from .rest_client import AtlassianRestAPI
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


rlmaers:
Looks like the CI failed due to checking with Python 2. Even if I'd change the f-string that it complains about, it would also fail on the exception chaining that's used extensively to keep the traceback information from the previous error(s).

gonchik:
Yeah, you are right, better adopt instead of find another way

PR #448, opened 2020-02-12
Divide examples into individual folders + another bamboo example

djachkov:
I think that is a good idea


gonchik:
Sounds good approach

PR #459, opened 2020-02-23
[Bamboo] Add disable_plan method

srosenthal:
I used this method to batch-disable several Bamboo branches I didn't want running anymore.

The relevant docs on the Atlassian side are at https://docs.atlassian.com/atlassian-bamboo/REST/6.10.4/#d2e4780, which shows the `/plan/{projectKey}-{buildKey}/enable` URL supports both POST to enable a plan and DELETE to disable a plan.

Opportunistic cleanup: the title of the Bamboo docs incorrectly said it was Confluence.


gonchik:
Thank you!

PR #460, opened 2020-02-24
[Confluence] Fix page title comparison

chdwlch:
This is a fix for a bug that makes the Confluence library think that every page needs to be updated.
The 'title' key does not exist in the dictionary returned by get_page_by_id() with expand set to 'body.storage', so current_title in the title comparison always evaluates to None.

I moved the title check before the body check and removed the expanded option from get_page_by_id so we can access the correct title value.

Tested against Confluence 6.13.8


gonchik:
Good caught! 
Thank you for your help!

PR #461, opened 2020-02-24
[Confluence] Remove faulty parent ID selection

chdwlch:
The parent content ID is NOT guaranteed to be the last element of the expanded ancestors list of a page. In my testing, the order of the ancestors is seemingly random. There is perhaps something in the database that is ordering the ancestors after an update occurs to the page.

This bug results in pages being updated and moved up the ancestor hierarchy randomly so as to prevent me from maintaining a hierarchical set of wiki pages with automated updates.

Tested against Confluence 6.13.8

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -86,16 +86,6 @@ class Confluence(AtlassianRestAPI):
         """
         return (self.get_page_by_title(space, title) or {}).get('id')
 
-    def get_parent_content_id(self, page_id):
-        """
-        Provide parent content id from page id
-        :type page_id: str
-        :return:
-        """
-        parent_content_id = ((self.get_page_by_id(page_id=page_id, expand='ancestors').get('ancestors') or {})[-1].get(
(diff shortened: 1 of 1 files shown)


chdwlch:
Good point. There might be another way to implement parent ID selection. 
First get the page's list of ancestors.
If no ancestors, the page is a root page.
If one ancestor, use that as parent.
If more than one ancestor, loop through each ancestor and find which one has the page listed as a child.

Can be used in update_or_create() function to validate the parent_id argument passed if the child page exists and is being updated.

gonchik:
Could you reproduce it after rebuilding the ancestor table, please? 
https://confluence.atlassian.com/conf612/rebuilding-the-ancestor-table-958778330.html

gonchik:
as I see that condition will cover your request https://github.com/atlassian-api/atlassian-python-api/commit/7af698e9438842a1f8ce925dc30f252fc77bb24d

PR #462, opened 2020-02-26
support proxy connection for rest client

er28-0652:
I just simply added `proxies` parameter for initialization of AtlassianRestAPI class. 
this parameter will be used on sending request.
So that we can use this awesome library even in proxy environment.


gonchik:
Got it, thanks!

Jmontes7:
Awesome, thank you very much!

PR #466, opened 2020-03-03
Check for if an element is a valid email

yhazrat:
The current` is_email()` has a faulty check for whether the input argument is an email. It accepts non string arguments but matches the regex only for arguments that are string. For all other argument types ex: integer, the check is returned True. The fix is to cast the argument as `str` and then perform the check


gonchik:
Thank you!

Issue #475, opened 2020-03-07
[Bitbucket] Pipelines REST API wrapper


pylipp:
Are there plans to wrap the [pipelines API](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pipelines/) or parts of it at some point? Would you accept contributions?

gonchik:
Hi!
Yeap, it sounds good and time to implement those rest end points.
Feel free to provide a PR, please.

Cheers, 
Gonchik

pylipp:
Fixed in #477

PR #477, opened 2020-03-12
Wrappers for Bitbucket Pipelines REST API

pylipp:
- originated from #475
- implementation of several Pipelines-related REST APIs
- tested with a small [test repo](https://bitbucket.org/pylipp/test-rest-api-wrapper/addon/pipelines/home#!/) of mine

Side note: I'm used to work with URLs like `https://api.bitbucket.org/2.0/user/repo/pipelines/`, hence I used the `resource_url()` method to construct URLs (I set up a Bitbucket instance as `bb = Bitbucket("https://api.bitbucket.org", api_version="2.0", api_root="")`. I'd find it beneficial if more methods of the `Bitbucket` class used this approach instead of having the `rest/api/` path hardcoded. I'd provide a separate PR if you don't see any objections.

### DIFF
--- a/atlassian/bitbucket.py
+++ b/atlassian/bitbucket.py
@@ -2272,3 +2272,40 @@ class Bitbucket(AtlassianRestAPI):
             version=version, workspace=workspace, repository=repository)
         return self.get(url, params={"pagelen": number, "sort": sort_by},
                         trailing=True)
+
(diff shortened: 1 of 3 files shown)


dtchanpura:
That's a good suggestion to make use of api_root instead of hard-coding `/rest/api` in every request and I would like to see that. We wanted to make use of these APIs and instead of using a client like this we end up using HTTP API to automate stuff at our end.

PR #480, opened 2020-03-19
Bitbucket. Added function to request PR participants

devopsvalkyrie:
Added a function to request participants and their statuses. It may be useful if you want to check whether a pull request was approved by any user or by a particular user. 
I tried to match code style around :)


gonchik:
Thanks

PR #481, opened 2020-03-21
[Bitbucket] Unification and url generate cleanup [outdated]

fedordikarev:
I think it could be good idea to remove `if not self.cloud` statements and use property for that and special methods like `self._get_api()` and `self._put_api()` to make api calls.
Here I replace almost all `self.put` for example. What do you think?
Should I continue with that changes?

### DIFF
--- a/atlassian/bitbucket.py
+++ b/atlassian/bitbucket.py
@@ -8,6 +8,23 @@ log = logging.getLogger(__name__)
 
 class Bitbucket(AtlassianRestAPI):
     bulk_headers = {"Content-Type": "application/vnd.atl.bitbucket.bulk+json"}
+    api_base_url = None
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


pylipp:
I think the URL generation could take some refactoring, so this PR is a good start.
However I'd make use of the `api_root` and `api_version` attributes and the `resource_url()` method of `AtlassianRestAPI` instead of introducing new methods. I used this method in #477.

gonchik:
I just thinking about that parameters as well.
https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/rest_client.py#L24

Spacetown:
Oops, I did also a refactoring with #609 which was merged today.
Currently I'm working on a object oriented implementation of the API.

gonchik:
Closing as it's confusing

PR #490, opened 2020-04-20
Jira8 set deleted as unnecessary

SLRover:
All improvements will implementing in just Jira set


gonchik:
@SLRover  Do you think it's time to remove that?

gonchik:
time to remove that

PR #494, opened 2020-04-27
Add JQL queries to service_desk.py

sekritskwurl:
Adding the ability in service desk to perform JQL queries. 
This was taken directly from the jira.py module and placed so that it could be called directly from the service desk module. 
Seems to work great from preliminary testing.


gonchik:
Hi!
Looks like better to make a inheritance code.

Issue #495, opened 2020-04-27
[Jira] transition with fields


suhalo:
In our Jira environment we have transitions that need mandatory fields. These fields have to be set during transition. 
See [atlassian doc](https://docs.atlassian.com/software/jira/docs/api/REST/8.5.4/#api/2/issue-doTransition)
Looking trough the implemented transition methods I could not find one having the ability to add data fields to the transition.

PR #496, opened 2020-04-27
Add fields to issue transitions

suhalo:
As described in  #495 this adds support for transitions with mandatory fields.
The default behavior of `set_issue_status` is not changed as the new parameter `fields` defaults to None


gonchik:
Thanks!

PR #519, opened 2020-06-15
Add ability to set a label on a repository via the API

aregier:
This will come as a huge surprise, but after being able to read repository labels, it would be nice to be able to set them as well :-)


gonchik:
Thank you! :)

PR #530, opened 2020-07-06
Feature/api client post

satyamsoni2211:
- Enhanced APIClient to include JSON parameters in post request


satyamsoni2211:
I have opened this pull request to directly post data as JSON instead of converting the data into JSON encoded string and passing in post API. I am using the same patch in my existing code in my organization at HSBC.

gonchik:
Hi!
Unfortunately, to me is unclear the purpose of wrapping into json field for request method. 
As it's not cover for all other requests

gonchik:
Do you have plan to provide a PR with all needful changes?

satyamsoni2211:
The purpose of using JSON parameters is to pass dict of dict as **JSON** directly to request body for the creation of JSD ticket. I was facing issues with post requests for multiple hierarchies of dict for custom fields containing sub child fields and options. User would now be able to use data/JSON parameters, any of them as per there convenience.

satyamsoni2211:
I have updated PR to be backward compatible, now user can either use JSON encoded values_dict or pass it as is, the argument will be chosen based on the type of payload.

gonchik:
@satyamsoni2211  Thank you for your update. 
Actually,  my question was about the data transferring to python-requests method.
As you updated the parameters for request 
```
request(self, method='POST', path='/', data=None, flags=None, params=None, json=None, headers=None,
                files=None, trailing=None)
```
but here that json will not transfer.
```
response = self._session.request(
            method=method,
            url=url,
            headers=headers,
            data=data,
            timeout=self.timeout,
            verify=self.verify_ssl,
            files=files,
            proxies=self.proxies
        )

```

based on that, could you provide your main pain about setting the json? Maybe to make wrapper for all type of requets (GET, POST , PUT etc )

satyamsoni2211:
I will check this out and make the required changes

satyamsoni2211:
For future versions, I will make changes to all kinds of wrapper to accommodate the changes. Since I faced this issue with create request, I added this to the specific method, since now request method can accept JSON, I will test other methods out and can make the similar changes later on.

gonchik:
thanks

PR #531, opened 2020-07-10
Confluence add_space_permission method

darryllee:
So, I copied other methods, and cribbed directly from the API example (https://developer.atlassian.com/cloud/confluence/rest/api-group-experimental/#api-api-space-spacekey-permission-post) to get this to work.

Caveat - There's no error handling, so if, for example, the permission already exists for space, you get an ugly (and uninformative):
> 400 Client Error: Bad Request for url

I was just happy I got it to work at all. :-}

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -1965,7 +1965,6 @@ class Confluence(AtlassianRestAPI):
         """
         url = 'rest/api/space/{}/permission'.format(space_key)
         data = {
-            'id': 2154,
             'subject': {'type': subject_type, 'identifier': subject_id},
             'operation': {'key': operation_key, 'target': operation_target},
             '_links': {}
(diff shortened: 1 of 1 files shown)


darryllee:
Oh also, I only tested this with Cloud.

atlassian/confluence.py
```diff
@@ -1952,6 +1952,27 @@ def add_user_to_group(self, username, group_name):
...
+        :param space_key: str
+        :param subject_type: str
+        :param subject_id: str
+        :param operation_key: str
+        :param operation_target: str
+        :return: Current permissions of space
+        """
+        url = 'rest/api/space/{}/permission'.format(space_key)
+        data = {
+            'id': 2154,
```
gonchik, line 1968: Hi!
looks like id is hardcoded :)

darryllee: Yes, at one point when debugging (I think the problem was I was unnecessarily wrapping the data in extra layer of JSON), I put the id from the API example in.

I just tested with it removed, and it worked, so let me fix that.

gonchik:
Thank you :)

jsoncz:
I get null for uri within Server edition :( did anyone get this working on other than cloud?

gonchik:
@jsoncz  which version ?

jsoncz:
> @jsoncz which version ?

7.13.12 DC

PR #539, opened 2020-07-20
Add methods for notification schemes

VasiliyGrizel:
Add to methods. One for find all notification schemes and second for get all information about single scheme.


gonchik:
Hm, thank you for that pr. As I see I made a some mistake for other method

Issue #544, opened 2020-07-23
Confuence: Migration picture failed!


99Kies:
body = self.confluence.get_page_by_id(page_id, "body.storage", status=None ,version=None)

```
{'storage': {'value': '<p><ac:image ac:width="1000"><ri:attachment ri:filename="DCE.PNG" /></ac:image></p>', 'representation': 'storage', '_expandable': {'content': '/rest/api/content/3998809'}}, '_expandable': {'editor': '', 'view': '', 'export_view': '', 'styled_view': '', 'anonymous_export_view': ''}}
```

![image](https://user-images.githubusercontent.com/45918704/88255123-1f6ef380-ccea-11ea-9e6d-7aada8d6fceb.png)

gonchik:
@99Kies  I hope you are  upload DCE.PNG on page. Could you double check, pleasE?

99Kies:
> @99Kies I hope you are upload DCE.PNG on page. Could you double check, pleasE?

I'm working on a document migration. But the migration failed.
create_page()  body.storage...

99Kies:
https://github.com/99Kies/confluence_func
ok, speak with code

99Kies:
![image](https://user-images.githubusercontent.com/45918704/88364412-670e8180-cdb5-11ea-9b8a-f9153987f5c9.png)
I don't know why creating a page in Storage format causes image creation to fail.
So you mean this picture is attached?

gonchik:
You need to put attachment. Then add into macro as a link or other way.
Please, take a look 2 examples
https://github.com/atlassian-api/atlassian-python-api/blob/master/examples/confluence/confluence_attach_file_as_link.py
https://github.com/atlassian-api/atlassian-python-api/blob/master/examples/confluence/confluence_attach_file.py

PR #545, opened 2020-07-23
Add get_parent_content_title func

99Kies:
Page_id becomes useless when you implement article transfer between multiple web services.


99Kies:
#544 hey, guys!

PR #548, opened 2020-07-31
[Confluence] Return full response in Confluence.get_page_child_by_type

thinusbaun:
get_page_child_by_type should return whole response from Confluence, to be able to properly use pagination.

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -83,7 +83,7 @@ class Confluence(AtlassianRestAPI):
 
             raise
 
-        return response.get('results')
+        return response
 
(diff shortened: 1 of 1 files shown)


gonchik:
Hi @thinusbaun ,
Thanks for your PR, could you describe your use case, please?

Cheers,
Gonchik

thinusbaun:
In my application i use get_page_child_by_type to gather list of attachments for given page. There are some pages which have few hundred of attachments linked. In case when there is many attachments, i need to use pagination metadata to fetch whole list of attachments in chunks.

PR #550, opened 2020-08-07
BugFix: positional args start_index and max_results misplased in plans()

monester:
Make start_index and max_results args as keyword arguments.


gonchik:
Thank you!

PR #556, opened 2020-08-20
Fix cookie auth parameter in docs

modelrockettier:
The docs currently say to use the "cookie" parameter for cookie authentication, but the code actually uses the "cookies" parameter (plural).

See https://github.com/atlassian-api/atlassian-python-api/blob/23370dd1cbf7b1999131412096d5e1b4908098b8/atlassian/rest_client.py#L23-L25


gonchik:
Thank you! :) 
That's great to see all polishes

PR #559, opened 2020-08-26
feat: add xray rest api (server/data-center) support

KevinDeJong-TomTom:
Add support for v1.0 of the Xray REST api for Server and Data-Center. 
Xray is a Jira plugin which provides Test Management support. I use the REST api for test automation (+feedback) from CI.

NOTE:
* As of now, v2.0 does not cover all functionality provided in the initial version, mainly related to test automation (eg. import of test executions).
* The REST api of Xray for Jira Cloud differs from Jira Server/DC.


gonchik:
So let's make a release with that wrapper.

Issue #557, opened 2020-08-25
ServiceDesk - attach_temporary_file


kilsenp:
For me, the attachTemporaryFile call is failing. 
Not sure if this is related to my specific server setting, but in the experimentalHeader the content type is set to json, which causes an error to me (I think it was 415).

It can be resolved by just not setting a content type, which is done automatically by the request module when passing a file.

I have a fix in place by just creating the headers in place. If you want, I can create a PR for that.

gonchik:
Hi! 
Feel free to raise a PR

PR #560, opened 2020-08-28
Set correct header for attach_temporary_file for ServiceDesk

kilsenp:
Fixes: #557 

More like a hotfix...
https://docs.atlassian.com/jira-servicedesk/REST/3.6.2/#servicedeskapi/servicedesk/{serviceDeskId}/attachTemporaryFile-attachTemporaryFile
`This resource expects a multipart post`. 
The experimental headers are however set to `application/json`.


gonchik:
Thanks!

PR #577, opened 2020-09-03
Atlassian Rest Client: Enhancing Implementation

bl4ckst0ne:
Hey, 
I've thought we can add 2 essential things to the atlassian rest client, for enhancing the implementation.

## Context Managers

The internal session (`requests.Session`) should be closed after its usage. When closing the session, its adapters are closed. It is a good practice to close the session after we finish using it. 
Therefore, I've added the `close` method, for supporting closing the client, which will close the internal session of the client (for those who care about that).

In order to make it idiomatic, we can support the context mangers API, so that the following code will be valid:
```python
with Jira(url, username, password) as jira:
    # use jira client
```
Thanks to the context managers API, the session will be closed at the end of the block. In order to support it, we should implement the magic methods `__enter__` and `__exit__` for entering the block, and exiting from it.

## Improving Request Wrappers Methods

The methods that wrap the request methods, `post`, `put` and `delete`, repeat themselves. All of them try to get `response.json()` and if there is an exception, handle it. We can improve it, by creating a new method that will handle it for us.

**Note**: also, we can use a decorator, for handling exceptions of the requests (maybe to give parameters for which exceptions should be handled for the wrapped function). However, it will not be neat, due to the logging after handling the exceptions.

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -54,6 +54,12 @@ class AtlassianRestAPI(object):
         elif cookies is not None:
             self._session.cookies.update(cookies)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *_):
+        self.close()
+
     def _create_basic_session(self, username, password):
         self._session.auth = (username, password)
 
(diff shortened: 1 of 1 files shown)


gonchik:
Hi,
First of all, thank you for you PR. 
I've checked and read the code. It's pretty correct and useful changes for us. 
Also, as I see it's good way in the provide info about that into README.md and docs and examples.

bl4ckst0ne:
2 Questions:
1. Do you want to change the whole docs to use the `with` statement, or at least call the `close` method?
2. Do you want me to do it? If so, do you want me to do it in a new branch?

gonchik:
1. I think for docs enough to show close method. Regarding `with` statement nice to see in Readme.MD and and read in the some examples.
2. Yeap, why not :). Feel free to make it in forked branch and then send the pull request. 
Thank you for your questions

PR #584, opened 2020-09-08
[RestClient] Cleanup base class and fix init order in derived class

Spacetown:
Add `__init__` to Confluence and move specific code from `rest_client.py` to this new function.
Change order in `__init__` of Bitbucket to prevent errors if `__init__`of `AtlassianRestAPI` is dependent from the `cloud` flag.

### DIFF
--- a/atlassian/bitbucket.py
+++ b/atlassian/bitbucket.py
@@ -10,11 +10,12 @@ log = logging.getLogger(__name__)
 class Bitbucket(AtlassianRestAPI):
     bulk_headers = {"Content-Type": "application/vnd.atl.bitbucket.bulk+json"}
 
-    def __init__(self, *args, **kwargs):
-        super(Bitbucket, self).__init__(*args, **kwargs)
-        url = kwargs.pop('url', False)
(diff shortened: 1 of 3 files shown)


atlassian/bitbucket.py
```diff
@@ -10,11 +10,12 @@
 class Bitbucket(AtlassianRestAPI):
     bulk_headers = {"Content-Type": "application/vnd.atl.bitbucket.bulk+json"}
 
-    def __init__(self, *args, **kwargs):
-        super(Bitbucket, self).__init__(*args, **kwargs)
-        url = kwargs.pop('url', False)
-        if url and 'bitbucket.org' in url:
-            self.cloud = True
+    def __init__(self, **kwargs):
```
gonchik, line 13: Do we really need to skip the *args ? 
https://www.python.org/dev/peps/pep-0362/#boundarguments-object

Spacetown: There are no args, only kwargs.

gonchik:
Thank you! That's great!

PR #586, opened 2020-09-10
Jira: Added cluster methods

SLRover:
Added two new methods:

- delete_cluster_node
- set_node_to_offline


gonchik:
Thanks for fixing that one #569

PR #588, opened 2020-09-14
[Bitbucket] Enforce PR limit when using server

nafeger:
When pulling PRs from bitbucket server, the limit url parameter doesn't
seem to be beeing heeded. This change allows the library to prevent too
many PRs from being returned to the client.

In my particular version of bitbucket: Atlassian Bitbucket v5.14.1 
When calling this function I was getting back, even when called with a limit explicitly set. This change allows the client to help the server in the event the server returns more data than we want.

### DIFF
--- a/atlassian/bitbucket.py
+++ b/atlassian/bitbucket.py
@@ -801,7 +801,7 @@ class Bitbucket(AtlassianRestAPI):
                 pr_list.extend(response.get("values", []))
 
         else:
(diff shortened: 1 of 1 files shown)


gonchik:
Looks like it's make sense. 
Thank you!

nafeger:
Appreciate it. Gotten a lot of value out of this lib, feels nice to give back a little

PR #592, opened 2020-10-02
[ServiceDesk] Support multiple attached files for a single comment

kilsenp:
I would like to attach multiple files to an issue without spamming the comment section.
The change proposes to use a list instead of a string, but has some backwards compability.

As far as I know, there are better checks for iterables (while excluding strings). If necessary I can add that (or is there somewhere already a function that can do that?).

Best regards


gonchik:
@kilsenp  Can we wrap via new method name? 
as I see some scripts will have a problem after merging that data

PR #609, opened 2020-10-23
[BitBucket]: Refactor url handling.

Spacetown:
- Refactor API endpoints to reuse the common leading parts.
- Refactor get call for paging:
  - Move code for paging to a separate function and reuse it in the paged endpoints.
  - In advanced mode the raw data is returned,
  - If limit is given the next page is read until the given limit.
  - If limit is None the whole list is retrieved.

The functions are a reordered according to the endpoint url.


gonchik:
Hi! 
Thank you for that initiative!

PR #611, opened 2020-10-29
Fix broken Bitbucket Cloud client

mandrean:
This PR fixes a bunch of issues/regressions introduced in the BB Cloud client with #609.

**Note:** It seems like a lot of refactoring was done without a proper Python IDE? Something like [PyCharm](https://www.jetbrains.com/pycharm/) + [Mypy](https://plugins.jetbrains.com/plugin/11086-mypy) + [Pylint](https://plugins.jetbrains.com/plugin/11084-pylint) + [SonarLint](https://plugins.jetbrains.com/plugin/7973-sonarlint) would easily have alerted the developer of.

The PR also adds support for the alternative Bitbucket Cloud API URL, i.e. `https://api.bitbucket.org/2.0` instead of `https://bitbucket.org/api/2.0`.

PR #612, opened 2020-10-31
Add linting

mandrean:
This PR is mainly about adding a bunch of linters (flake8/pylint/black/mypy/bandit/doc8) to tox, and fixing the warnings/errors prompted by them.

It also adds two missing return statements that wasn't added in #611.

**NOTE:** `pylint` isn't actually activated in `[tox]envlist` yet because I ran out of time to fixk all those errors. Can be done in a later PR.


gonchik:
Thank you! :) 
Just curious, the single-quotes vs double-quotes, is this more the google do you know about that ?

Also, what do you think about the Github CI instead of travis?

mandrean:
@gonchik You're welcome! Very good questions:

1. About quotes, all _PEP_ says is ["be consistent"](https://www.python.org/dev/peps/pep-0008/#string-quotes), i.e. don't use singles sometimes and doubles sometimes. But [black](https://github.com/psf/black) uses doubles, so it's easy to just go with that IMHO? You can just run `$ black atlassian-python-api/` and you don't have to bother with the rest. Just like with `gofmt`/`rustfmt` in Go/Rust. There's a debate over here about why `black` doesn't have an option for singles: https://github.com/psf/black/issues/118

2. You're reading my mind! I was thinking about suggesting a move to GitHub CI in a later issue/ticket... I've used it on a few open- & closed-source projects, and it's really solid. It's gaining a lot of popularity with many projects. It's very fast, I feel like Travis is slower to boot up the different VMs/environments

PR #613, opened 2020-11-01
[Bitbucket] First approach for object oriented interface for bitbucket cloud.

### LINKED ISSUE #610: [BitBucket]: Suggestion for object oriented interface
My idea is to use a object oriented interface to the API points instead of using functions with many arguments. I've started to work on this on May fork.

All paged APIs shall return a generator object and the paging is down internally, removing all start and limit arguments from the user.
There will also be a single definition of the rest endpoint. 

In the first step I'll skip the outdated cloud entpoints. I also saw that a new function `get_repositories` was introduced instead of updating the existing function. I think that the cloud and the server should be splitted because of the different API structure. 

I've not checked if the current state is runnable. 

What do you think, shall I proceed?

https://github.com/Spacetown/atlassian-python-api/tree/bitbucket-object-oriented

Spacetown:
First apporach of implementing a OO interface as mentioned in #610.

The default reviewers are only readable. I use the same credentionals as in the web interface where I can create a default reviewer.

closes #610

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -1,6 +1,8 @@
 # coding=utf-8
 import logging
 
+from deprecated import deprecated
+
 from .base import BitbucketBase
 from atlassian.bitbucket.cloud import Cloud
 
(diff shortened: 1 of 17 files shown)


Spacetown:
@gonchik Shall I also add a new example or modify an existing one? Shall the old functions be marked as deprecated?

PR #622, opened 2020-11-08
[Bitbucket] Fix wrong URLs after refactoring

### LINKED ISSUE #619: [Bitbucket] Tag deletion does not work
Hello,

Using the `delete_tag` method against a Bitbucket server instance, I got an error 405.
Checking the [documentation](https://docs.atlassian.com/bitbucket-server/rest/7.7.1/bitbucket-git-rest.html), it appears that tag deletion goes through the Bitbucket **git** API, so the path should be `rest/git` instead of `rest/api`.

Digging a bit the source code, i think that we would need to add the `api_root` parameter on the methods and sub-methods  invoked by the 
`delete_tag` one. Would that be the right fix?

Thanks

Spacetown:
The commit a9c6159bcf98992a4bb892a2fb4ad8090c3bd836 changed the URL handling. The non default API roots are changed wrong. 
I've searched in the old implementation all endpoint not matching the default root `rest/api` with the regular expression `['"]\/?rest\/(?!api)` and checked the new implementation.

I think that the old implementation of `get_project_condition` used the wrong endpoint. According to the [documentation](https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-default-reviewers-rest.html#idm52264901504) the endpoint for a specific condition is `/rest/default-reviewers/1.0/projects/{projectKey}/condition/{id}` and not `.../conditions/{id}`.

closes #619

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -87,7 +87,7 @@ class Bitbucket(BitbucketBase):
         Rebuild the bundled Elasticsearch indexes for Bitbucket Server
         :return:
         """
-        url = self.resource_url("sync", api_root="indexing", api_version="latest")
+        url = self.resource_url("sync", api_root="rest/indexing", api_version="latest")
         return self.post(url)
 
     def check_reindexing_status(self):
@@ -95,7 +95,7 @@ class Bitbucket(BitbucketBase):
         Check reindexing status
         :return:
         """
-        url = self.resource_url("status", api_root="indexing", api_version="latest")
+        url = self.resource_url("status", api_root="rest/indexing", api_version="latest")
         return self.get(url)
 
     def get_users(self, user_filter=None):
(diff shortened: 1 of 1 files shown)


atlassian/bitbucket/__init__.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, url, *args, **kwargs):
                 "2.0" if "cloud" in kwargs and kwargs["cloud"] else "1.0"
             )
         if "cloud" in kwargs:
-            kwargs["api_root"] = "" if "api.bitbucket.org" in url else "api"
+            kwargs["api_root"] = "" if "api.bitbucket.org" in url else "rest/api"
```
mandrean, line 23: This is incorrect, the `api_root` for Bitbucket **Cloud** is either `api.bitbucket.org` or `bitbucket.org/api`, there's no `/rest` in there

mandrean (changes requested):

atlassian/bitbucket/__init__.py
Spacetown: Of course there should no `rest` be added.

PR #623, opened 2020-11-10
[BitBucket] Add mockup for the URL response in the test of bitbucket.

Spacetown:
This is a first approach to add tests for the CI without connecting to a bitbucket instance.

@gonchik Shall I continue with the rest of the cloud functions?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ __pycache__/
 *.py[cod]
 *$py.class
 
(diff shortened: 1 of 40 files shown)


gonchik:
@Spacetown  Is it ready to check ?

Spacetown:
@gonchik Yes, it's ready. The test will be extended in the feature with all of the properties for the OO interface.

PR #624, opened 2020-11-12
[BitBucket]: fixing work with comments

alexander-gluhov:
…159bcf98992a4bb892a2fb4ad8090c3bd836

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -1437,8 +1437,8 @@ class Bitbucket(BitbucketBase):
     def _url_pull_request_comment(
         self, project_key, repository_slug, pull_request_id, comment_id
     ):
-        url = "{}/comments/{}".format(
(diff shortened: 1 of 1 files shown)


atlassian/bitbucket/__init__.py
```diff
@@ -1437,7 +1437,7 @@ def add_pull_request_comment(
     def _url_pull_request_comment(
         self, project_key, repository_slug, pull_request_id, comment_id
     ):
-        url = "{}/{}".format(
+        url = "{}/comments/{}".format(
             self._url_pull_request(project_key, repository_slug, pull_request_id),
```
Spacetown, line 1441: Instead of adding `comments` the `_url_pull_request`should be replaced with `_url_pull_request_comments`. I've added this fix also to #622 .
```suggestion
        url = "{}/{}".format(
            self._url_pull_request_comments(project_key, repository_slug, pull_request_id),
```

Spacetown (changes requested):

PR #625, opened 2020-11-14
There is no status method in the response message

monkey-wenjun:
There is no status method in the response message


gonchik:
Thanks it's useful

PR #627, opened 2020-11-15
[Tests] Remove Travis CI

Spacetown:
Extend GitHub actisons with python3.9 and python-dev (apt-get doesn't know 3.10-dev). Need to switch to Upuntu 20.04 and gcc-8 to use python3.9.

@gonchik please extend the required tests.


Spacetown:
@gonchik The reverting to 61729af8 in 188d8945 wasn't succesful. I did a Cerry-Pick of 61729af8.

gonchik:
Thank you for your fix and contribution!

PR #632, opened 2020-11-24
WIP: [Confluence] : Update to support Cloud REST API endpoints

korrosivesec:
Used the bitbucket module as an example and updated confluence to support cloud endpoints.


Spacetown:
The base classes are created for a OO interface. You're not using a OO interface therfore it's very confusing.

korrosivesec:
Ok, I made this a Work In Progress for now and will take a closer look and try to clean things up.

PR #634, opened 2020-11-25
[ServiceDesk] Added the "Create request type" method available on the REST API "requestType" endpoint

dbarroso:
The _requestType_ API endpoint has a _createRequestType_ method that can be used to create new request types.

Documentation here:
https://docs.atlassian.com/jira-servicedesk/REST/4.13.1/#servicedeskapi/servicedesk/{serviceDeskId}/requesttype-createRequestType

I've added some code to implement this method.


Spacetown:
Can you use black to format the code or use the black output to update your code?

gonchik:
Thanks

PR #637, opened 2020-11-28
[Bitbucket] Fixing work with Bitbucket Server and one new method

alexander-gluhov:


gonchik:
Hi!
@alexander-gluhov  Thank you for that PR.

PR #640, opened 2020-11-30
[Bitbucket] Add OO interface for projects and repositories in bitbucket server.

Spacetown:
Add first OO interfaces with a test response for each function.
Validate size of paged response.
Extend VSCode workspace with file association for the responses.


gonchik:
Looks interesting

Spacetown:
But interesting is not good enough.

Spacetown:
Should I add more endpoints in this PR or do it in a separate one?

gonchik:
Let's do in separate one.
I will play with changes soon

Issue #641, opened 2020-12-07
bitbucket.set_branching_model broken for Bitbucket Server edition


aregier:
Sometime after 1.16.0, when calling the `set_branching_model` API against a Bitbucket Server, a 404 is returned. 
The URL for this api is not set correctly.

1.16.0: `https://myserver.local/bitbucket/rest/branch-utils/1.0/projects/{projectSlug}/repos/{repoSlug}/branchmodel/configuration`

Latest: `https://myserver.local/bitbucket/rest/branch-utils/1.0/rest/api/1.0/projects/{projectSlug}/repos/{repoSlug}/branchmodel/configuration`

Spacetown:
I'm checking this.

gonchik:
#642 thanks for the merge request

PR #642, opened 2020-12-07
Fix issue #641 (bitbucket.set_branching_model api)

aregier:


atlassian/bitbucket/__init__.py
```diff
@@ -1948,9 +1948,9 @@ def all_branches_permissions(self, project_key, permission_id, repository_slug=N
 
     def _url_branching_model(self, project_key, repository_slug):
         return self.resource_url(
-            "{}/branchmodel/configuration".format(self._url_repo(project_key, repository_slug)),
-            api_root="rest/branch-utils",
-            api_version="1.0",
+            "{}/branchmodel/configuration".format(self._url_repo(project_key, repository_slug, api_root="rest/branch-utils", api_version="1.0")),
+            api_root='',
+            api_version=''
         )
```
Spacetown, line 1954: Test fails because of formating. Following should work:
```python
        return "{}/branchmodel/configuration".format(
            self._url_repo(
                project_key,
                repository_slug,
                api_root="rest/branch-utils",
                api_version="1.0",
            )
        )
```

Spacetown (changes requested):

atlassian/bitbucket/__init__.py
aregier: @Spacetown - Better code, too. :-)

Spacetown (approved):
LGFM

gonchik:
Thank you!

PR #643, opened 2020-12-08
Bitbucket: Feature/Update Repo command

timsavage:
The Update Repo command is missing from the Bitbucket API allowing for updates to repository eg description. 

The method arguments follow the same pattern used by the `update_project` method.

The change covers both code and docs, and all tox checks pass.

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -305,7 +305,7 @@ class Bitbucket(BitbucketBase):
         """
         url = self._url_project(key)
         return self.put(url, data=params)
-    
+
     def _url_project_avatar(self, project_key):
         return "{}/avatar.png".format(self._url_project(project_key))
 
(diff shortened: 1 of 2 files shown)


atlassian/bitbucket/__init__.py
```diff
@@ -682,6 +682,23 @@ def get_repo(self, project_key, repository_slug):
...
 
+    def update_repo(self, project_key, repository_slug, **params):
+        """
+        Update a repository in a project. This operates based on slug not name which may
+        be confusing to some users.
+        :param project_key: Key of the project you wish to look in.
+        :param repository_slug: url-compatible repository identifier
+        :return: None if the repo does not exist, else the value of the put request.
+        """
+        data = self.get_repo(project_key, repository_slug)
```
Spacetown, line 693: The get isn't needed becore the put because missing Keys in the put aren't changed.

Spacetown (changes requested):

timsavage:
That is what I would expect from a put, however that was the pattern used
in the put request to update a project.

Will make the update.

Cheers

Spacetown:
@timsavage I was wondering when implementing the OO interface and checked it with BB cloud. It is working without the GEt.

timsavage:
Done, the update_project method now matches.

PR #645, opened 2020-12-10
Add URL encoding for multiple GET parameters

dim0x69:
Disclaimer: I only looked at Jira until now :)

Currently there seems not to be a method which adds multiple GET parameters to the API request. Indeed, these might happen in the future (I will add another Pull Request  to add the "expand" keyword to Jira.issue_field_value() and Jira.issue() ;) )

This Pull Request adds the option to have multiple GET parameters in the URL.

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -197,7 +197,7 @@ class AtlassianRestAPI(object):
         params_already_in_url = True if "?" in url else False
         if (params or flags) and not params_already_in_url:
             url += "?"
-        else:
+        if (params or flags) and params_already_in_url:
             url += "&"
         if params:
             url += urlencode(params or {})
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -195,10 +195,11 @@ class AtlassianRestAPI(object):
         """
         url = self.url_joiner(None if absolute else self.url, path, trailing)
         params_already_in_url = True if "?" in url else False
-        if (params or flags) and not params_already_in_url:
-            url += "?"
-        if (params or flags) and params_already_in_url:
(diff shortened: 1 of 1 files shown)


atlassian/rest_client.py
```diff
@@ -197,6 +197,8 @@ def request(
         params_already_in_url = True if "?" in url else False
         if (params or flags) and not params_already_in_url:
             url += "?"
+        else:
```
Spacetown: This should be:
``python
elsif params_already_in_url:
```
At the moment the `&`i s added even if tere are no `parmas`and no `flags`.

Spacetown (changes requested):

atlassian/rest_client.py
Spacetown: Ok. Instead of the branch I selected the first commit.

dim0x69:
Hi @Spacetown ,

Thanks for your suggestions! On first sight I (logically) don't see a difference in our codes, but yours is nicer to read :) 
I updated the Pull Request with  0fcc4d7 to include your idea.

Spacetown:
> Hi @Spacetown ,
> 
> Thanks for your suggestions! On first sight I (logically) don't see a difference in our codes, but yours is nicer to read :)
> I updated the Pull Request with [0fcc4d7](https://github.com/atlassian-api/atlassian-python-api/commit/0fcc4d7ee78fa907be5835015da9f87b2547dd30) to include your idea.

I've checked again and you're right, it was ok. But I think, now it's easier to read.

dim0x69:
I updated the PR with black formatting @Spacetown

gonchik:
Thank you both for the improvement!

Issue #454, opened 2020-02-19
ServiceDesk - add_users_to_organization - using user_list is deprecated


eos-13:
Hi,

According to Jira API documentation, using user_list for adding an user into organization is deprecated. Here is a suggestion of code to implement to goes around deprecation :+1: 

`
def add_users_to_organization_arrayId(self, organization_id, array_list):
    """
    Adds users to an organization
    users_list is a list of accountId

    :param self: object
    :param organization_id: str
    :param array_list: list
    :return:
    """
    logger.info('Adding users...')
    url = 'rest/servicedeskapi/organization/{}/user'.format(organization_id)
    data = {'accountIds': array_list}

    return self.post(url, headers=self.experimental_headers, data=data)
`

Best regards,

SLRover:
Hi, eos-13!

You can make Pull Request. It would be a good decision. 👍 
Or you can wait. 😉

PR #649, opened 2020-12-16
Update adding and removing users to organization

amdemas:
This covers #454 due to the deprecation noted [here](https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-user-privacy-api-migration-guide/).


gonchik:
Thanks!

Issue #652, opened 2020-12-20
[BITBUCKET CLOUD] Invalid pagination / missing next


sti0:
I tried to loop over my workspaces and over the repos to get all pipelines with the following code:

```
from atlassian.bitbucket.cloud import Cloud
from pprint import pprint

bb = Cloud(
    url="https://api.bitbucket.org",
    username="myusr",
    password="mypass",
)

for x in bb.workspaces.each():
    print(x.name)
    if x.slug == "myworkspace":
        y = x.repositories.get("myrepo")
        print(y.name, y.description)

        pipes = y.pipelines.each()
        for p in pipes:
            print(p.build_number)
```

It returns the following error
```
AssertionError: Wrong response for url [https://api.bitbucket.org/2.0/repositories/.../.../pipelines], the size attribute doesn't match the number of recieved values:
```

JSON Output of the request like this one:
```
{
    "page": 1,
    "pagelen": 10,
    "values": [....10 pipes here...],
    "size": 21
}
```
The `size` option gave me the total number of pipelines (21). But the response only gave me 10 pipes with no `next` attribute. Therefore an error happend here:

https://github.com/atlassian-api/atlassian-python-api/blob/4066a9e6da06a57fe411fb703c8b0acb17c88be1/atlassian/bitbucket/cloud/base.py#L62

Spacetown:
That was my fault. the size value is the overall size and not the size of the current response.

Spacetown:
One hint from my site. The workspace is accessed by the slug or the uuid. Therefore changing the code:
```python
for x in bb.workspaces.each():
    print(x.name)
    if x.slug == "myworkspace":
```
to 
```python
x = bb.workspaces.get("myworkspace")
```
will be faster if you have several workspaces.

sti0:
Thanks for the quick fix.

Another error occured when calling `completed_on` @ pipeline level:
```
return self.get_data("completed_on", "never completed")
TypeError: get_data() takes 2 positional arguments but 3 were given
```
https://github.com/atlassian-api/atlassian-python-api/blob/c51db61ae02333e9454a0ebdd0923f080be8f890/atlassian/bitbucket/cloud/repositories/pipelines.py#L120

sti0:
And according to the docs, `steps` should return a generator. But it returns a method.

```
for p in pipes:
            print(type(p.steps))
-----> <class 'method'>
```
[...]

PR #653, opened 2020-12-20
[Bitbucket] Fix AssertionError error in pagination.

Spacetown:
Fix assertion error from #652. The size value is the overall size and not the size of the current response. Therefore the assertion in the pagination is removed.


sti0 (approved):
Works as expected.

gonchik:
Awesome!

PR #655, opened 2020-12-20
[Bitbucket Cloud] Add basic pull request support

sti0:
This adds basic pull requests support for Bitbucket Cloud.

- [x] Get pull requests by repository
- [x] basic pull request informations
- [x] Participant support
- [x] Merge
- [x] Approve / Unapporve
- [x] Comment 
- [x] Decline
- [x] Request changes
- [x] Testing
- [x] Write unit tests


atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,169 @@
...
+        """ pull request description """
+        return self.get_data("description")
+
+    @property
+    def state(self):
+        """
+        pull request state
+        possible values: MERGED, SUPERSEDED, OPEN, DECLINED
+        """
+        return self.get_data("state")
```
Spacetown, line 74: Is it better to add a property for each state (`is_merged`, `is_superseded`, `is_open` or `is_declined`) instead of or in addition to returning the strings from BB?

atlassian/bitbucket/cloud/base.py
```diff
@@ -7,10 +7,11 @@
 
 
 class BitbucketCloudBase(BitbucketBase):
-    def __init__(self, url, *args, **kwargs):
+    def __init__(self, url, link="self", *args, **kwargs):
```
Spacetown, line 10: I'm wondering about this. Why is this needed?

atlassian/bitbucket/cloud/base.py
```diff
@@ -58,15 +59,7 @@ def _get_paged(self, url, params=None, data=None, flags=None, trailing=None, abs
...
-            values = response.get("values", [])
-            if not response.get("size", -1) == len(values):
-                raise AssertionError(
-                    "Wrong response for url [{}], the size attribute doesn't match the number of recieved values:\n{}".format(
-                        self.url, response
-                    )
-                )
-
-            for value in values:
+            for value in response.get("values", []):
```
Spacetown, line 62: There is #653 to fix this issue.

Spacetown (changes requested):

atlassian/bitbucket/cloud/repositories/pullRequests.py
sti0: sure, thats way better. Will add this..

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def declined_reason(self):
+        """ reason for declining """
+        return self.get_data("reason")
+
+    @property
+    def author(self):
+        """ DefaultReviewer object of the author """
+        return DefaultReviewer(None, self.get_data("author"))
```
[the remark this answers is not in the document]
sti0: It's the same structure but we should rename `DefaultReviewer` into `User`

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,169 @@
...
+    def state(self):
+        """ Returns approved, changes_requested or None"""
+        return self.get_data("state")
+
+    @property
+    def participated_on(self):
+        """ time of last participation """
+        po_str = self.get_data("participated_on")
+        po_dt = datetime.strptime(po_str, "%Y-%m-%dT%H:%M:%S.%f%z") if po_str else po_str
+        return po_dt
```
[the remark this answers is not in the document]
sti0: I thought returning a `datetime` object is more comfortable. In the most cases you would do this convention yourself later. 

[It like implementing `state` vs `is_open`, `is_closed` etc. ](https://github.com/atlassian-api/atlassian-python-api/pull/655#discussion_r546903402)

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,169 @@
...
+    def created_on(self):
+        """ time of creation """
+        return datetime.strptime(self.get_data("created_on"), "%Y-%m-%dT%H:%M:%S.%f%z")
+
+    @property
+    def updated_on(self):
+        """ time of last update """
+        uo_str = self.get_data("updated_on")
+        uo_dt = datetime.strptime(uo_str, "%Y-%m-%dT%H:%M:%S.%f%z") if uo_str else uo_str
+        return uo_dt
```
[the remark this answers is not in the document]
sti0: This is needed because `updated_on` maybe None which can't be handled by a `datetime` object.

atlassian/bitbucket/cloud/base.py
sti0: This is needed because the participants have no `link` attribute.
```
"participants": [
        {
            "participated_on": None,
            "state": None,
            "role": "REVIEWER",
            "user": {
                "display_name": "DefaultReviewer1",
                "uuid": "{DefaultReviewer1Uuid}",
                "links": {
                    "self": {"href": "bitbucket/cloud/2.0/users/%7BDefaultReviewer1Uuid%7D"},
                    "html": {"href": "https://bitbucket.org/%7BDefaultReviewer1Uuid%7D/"},
                    "avatar": {
                        "href": "https://secure.gravatar.com/avatar/31a164e6406a8deeb56e94c778030a9f?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FS-3.png"
                    },
                },
                "nickname": "DefaultReviewer1Nickname",
                "type": "user",
                "account_id": "DefaultReviewer1AccountId",
            },
            "type": "participant",
            "approved": False
        },
        {
            "participated_on": "2020-12-20T12:11:57.822940+00:00",
            "state": None,
            "role": "PARTICIPANT",
            "user": {
                "display_name": "DefaultReviewer2",
                "uuid": "{DefaultReviewer2Uuid}",
                "links": {
                    "self": {"href": "bitbucket/cloud/2.0/users/%7BDefaultReviewer2Uuid%7D"},
                    "html": {"href": "https://bitbucket.org/%7BDefaultReviewer2Uuid%7D/"},
                    "avatar": {
                        "href": "https://secure.gravatar.com/avatar/31a164e6406a8deeb56e94c77803123?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FS-3.png"
                    },
                },
                "nickname": "DefaultReviewer2Nickname",
                "type": "user",
                "account_id": "DefaultReviewer2AccountId",
            },
            "type": "participant",
            "approved": False
        }
    ],m
```

atlassian/bitbucket/cloud/base.py
sti0: Sure, had added this for my tests and get it to work. False commited.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: See https://github.com/atlassian-api/atlassian-python-api/pull/655#discussion_r546918654

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        pull request state
+        possible values: MERGED, SUPERSEDED, OPEN, DECLINED
+        """
+        return self.get_data("state")
+
+    @property
+    def created_on(self):
+        """ time of creation """
+        return datetime.strptime(self.get_data("created_on"), "%Y-%m-%dT%H:%M:%S.%f%z")
```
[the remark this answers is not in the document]
sti0: Same as https://github.com/atlassian-api/atlassian-python-api/pull/655#discussion_r546918654

atlassian/bitbucket/cloud/repositories/users.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Deletes the default reviewer
+        """
+        data = super(User, self).delete(None)
+        if "errors" in data:
+            return
+        return User(self.url, data, **self._new_session_args)
+
+
+class Participant(BitbucketCloudBase):
```
sti0, line 36: should we add a `is_default_reviewer` property?

atlassian/bitbucket/cloud/base.py
Spacetown: What about using following construct instead of the additional argument or to extend the `get_link` to check the existence of the key first and return None if it doesn't exists. 

```python
        if url is None:
            try:
                url = self.get_link("self")
            except KeyError:
                pass
```

sti0: What if it is needed but somehow not received. Than it would not fail and return None . This could lead to a problem elsewhere. In my opinion a value should only tried to be determined if it was expected.

atlassian/bitbucket/cloud/repositories/users.py
```diff
@@ -0,0 +1,75 @@
+from requests.models import ProtocolError
+from ..base import BitbucketCloudBase
+from datetime import datetime
+
+
+class User(BitbucketCloudBase):
```
Spacetown, line 6: I would move the `User` to  `atlassian/bitbucket/cloud/common/user.py` and use this as base for all other useers. E.G. The `delete` is implemented in `DefaultReviewer`, in the inherited class also other properties can be added.

atlassian/bitbucket/cloud/repositories/users.py
Spacetown: For the Participiant make sense.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: We should do this on the other classes as well

atlassian/bitbucket/cloud/repositories/users.py
sti0: Ok will do this. `Participant` will be a part of `pullRequests.py`

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
Spacetown: Ok. The already existing objects return the raw string and I don't like the hardcoded format. What about using a `timeformat` in the constructor with a default of None. The user can specify a format in the main constructor:
```python
cloud = Cloud(url="http://localhost:7990", username="admin", password="admin", timeformat="%Y-%m-%dT%H:%M:%S.%f%z")
```
The format must be given to the sub classes by extending the following function with the `timeformat`.
https://github.com/atlassian-api/atlassian-python-api/blob/eec2934fcd72ebf35205296bf2ddb3f94680c0ed/atlassian/bitbucket/base.py#L66-L72
In the base class we can define the function `get_time`:
```python
def get_time(self, id, default):
    if id not in self.__data:
        return default
   timestring = self.__data[id]
    if self.timeformat is not None:
        timestring = datetime.strptime(timestring, self.timeformat)
    return timestring
```

The function can than be used the smae way as `get_data`:
```python
    @property
    def completed_on(self):
        return self.get_time("completed_on", "never completed"))
```

Or maybe the `get_data`

atlassian/bitbucket/cloud/repositories/users.py
Spacetown: I alos suggest to remove the `delete` method from the `User`.

atlassian/bitbucket/cloud/base.py
Spacetown: I'm not sure if I understand your point correct. You mean that BB returns the correct response type but without the link to self? I think that's very unlikely.

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @property
+    def close_source_branch(self):
+        """ close source branch flag """
+        return self.get_data("close_source_branch")
+
+    @property
+    def source_branch(self):
+        """ source branch """
+        return self.get_data("source")["name"]["branch"]
```
Spacetown, line 114: Was my mistake.
```suggestion
        return self.get_data("source")["branch"]["name"]
```

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,197 @@
...
+    def reviewers(self):
+        """ Returns a generator object of reviewers """
+        for reviewer in self.get_data("reviewers"):
+            yield User(None, reviewer)
+
+        return
+
+    def comment(self, raw_message):
+        """ Commenting the pull request in raw format """
+        markupstrings = ["markdown", "creole", "plaintext"]
```
Spacetown, line 157: Unused variable.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: But `%Y-%m-%dT%H:%M:%S.%f%z` is the delivered Timeformat from BB.  It's necessary to use this format to convert `string` to `datetime`. Therefore it doesn't make sense to use custom set time format by the user. We can add a parameter to return raw timestamp or `datetime` object.

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,197 @@
...
+        self._check_if_open()
+        data = {"approved": True}
+        return self.post("approve", data)
+
+    def unapprove(self):
+        """ Unapporve a pull request if open """
+        self._check_if_open()
+        return self.delete("approve")
+
+    def merge(self, merge_strategy="merge_commit", close_source_branch=None):
```
Spacetown, line 180: In C we say 'No magic constants'. I prever class static variables for the string. E.g.
```python
class PullRequest(BitbucketCloudBase):
    MERGE_COMMIT = "merge_commit"
    SQUASH = "squash"
    FAST_FORWARD = "fast_forward"

    def merge(self, merge_strategy=PullRequest.MERGE_COMMIT, close_source_branch=None):
```
The string is defined in one place and you get a script error for a typo.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
Spacetown: Ok. My understanding was wrong. I thougt the function interprets the string and formats it according to the argument. The parameter to return the raw time is a good idea. Can this be combined with my wrong understanding of the function to give a format? E.g. Add a `timeformat` argument with a default of `""` which returns the `datetime` object, a value of `None` returns the raw string and other values are given to the function `strftime`of the `datetime` object.

atlassian/bitbucket/cloud/base.py
sti0: Yes. Maybe for technical reasons, bugs, incidents or whatever. This would not break.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: Yes, but i suggest only to impement a `raw_timeformat` boolean with default `False`. Convertings are user based. Most cases you need to convert to your timezone because BB send UTC time etc.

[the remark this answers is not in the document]
Spacetown: That's why I suggested a format string in the constructor. Than the user can define the format in a single place and don't have to take care about converting the time properties.

atlassian/bitbucket/cloud/repositories/pullRequests.py
sti0: Maybe we should generate a `const.py` to place all constants there..

Spacetown: But this can result in a big list and you need to comment where the constant is used.
For the permissions in the server implementation I used the local approach.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: But if you like to get a time string `DD.MM.YYYY HH24:MI` in GMT+2 you need to convert the timezone of the `datetime` object first and afterwards you could parse it to a string with the given format. Therefore you need a min. of two parameters (time format + timezone).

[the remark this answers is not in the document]
sti0: Added the possibility to define a `raw_timezone` boolean (default `False`). Maybe we could reconsider the suggestion about a pre-formatted time-string later. IMHO thats no breaking change on a later implementation.

atlassian/bitbucket/cloud/repositories/pullRequests.py
sti0: Sure, but thats the reason why we should do this. There is no need to define constants twice. Thats a pretty common way in other projects. We could even add prefix to the constants like PR_ for pull requests or PIPE_ for pipelines etc.

Nevertheless in such a case server should be separated from cloud.

atlassian/bitbucket/cloud/repositories/defaultReviewers.py
```diff
@@ -22,7 +23,7 @@ def add(self, user):
 
         :param user: string: The user to add
 
-        :return: The added DefaultReviewer object
+        :return: The added User object
```
sti0, line 26: TODO: DefaultReviewer object....

atlassian/bitbucket/cloud/repositories/defaultReviewers.py
```diff
@@ -70,31 +71,15 @@ def get(self, user):
...
-    def uuid(self):
-        return self.get_data("uuid")
+        super(DefaultReviewer, self).__init__(url, data, *args, **kwargs)
 
     def delete(self):
         """
         Deletes the default reviewer
         """
-        data = super(DefaultReviewer, self).delete(None)
+        data = super(User, self).delete(None)
```
sti0, line 82: TODO: DefaultReviewer

atlassian/bitbucket/cloud/repositories/defaultReviewers.py
```diff
@@ -70,31 +71,15 @@ def get(self, user):
...
         if "errors" in data:
             return
-        return DefaultReviewer(self.url, data, **self._new_session_args)
+        return User(self.url, data, **self._new_session_args)
```
sti0, line 85: TODO: DefaultReviewer

atlassian/bitbucket/cloud/repositories/pullRequests.py
sti0: `def merge(self, merge_strategy=PullRequest.MERGE_COMMIT, close_source_branch=None):` this would not work. You can't specify constants defined in a class as defaults. Therefore it must be defined outside of the class e.g. on top of the package or as mentioned in a `const.py`. 

That's another reason why we should use it.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
Spacetown: Later this will be a braking change or an additional argument is needed. What about configuring a function to change the time? None will return the raw string and the default `lamda x: x` will return the datetime object?

atlassian/bitbucket/cloud/repositories/pullRequests.py
Spacetown: But from the OO view the constants are part of the module and should not be defined in a global file. If I search the constants for a function I'm searching in the namespace/package of the function and not for a prefixed constant in a global package.
I prefer a package variable and a class variable referencing the package variable to reference the value with the dot notation of the Pull Request instance.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: Good catch. Added this in the last commit.

Default is convert to UTC `datetime`, if passing `None` returns the raw string. Otherwise you are able to pass a `lambda` function in.

```
import pendulum
from datetime import datetime
from atlassian.bitbucket.cloud import Cloud
from atlassian.bitbucket.cloud.const import CONF_TIMEFORMAT

def timeformat(x):
    tz = pendulum.timezone("Europe/Berlin")
    x = datetime.strptime(x, CONF_TIMEFORMAT)
    return tz.convert(x)


bb = Cloud(
    url="https://api.bitbucket.org",
    username="xxxx",
    password="xxxx",
    timeformat_func=lambda x: timeformat(x), # returns converted datetime object in timezone Europe/Berlin
    #timeformat_func=None, # -> returns raw string
)
```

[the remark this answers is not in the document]
sti0: @Spacetown what about the parameter name? Should it be `timeformat_func` or `timeformat_lambda` or do you have another idea.

I tend to `timeformat_lambda`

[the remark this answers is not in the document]
Spacetown: I think `timeformat_lambda` is fine.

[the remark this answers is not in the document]
Spacetown: I think the line `x = datetime.strptime(x, CONF_TIMEFORMAT)` should be done in `get_time`:
```python
    def get_time(self, id):
        value_str = self.get_data(id)
        if not value_str or not isinstance(value_str, str):
            return None
        else:
            return value_str if self.timeformat_func is None else self.timeformat_func(datetime.strptime(x, CONF_TIMEFORMAT))
```
Than the user doesn't need to know the time formet `CONF_TIMEFORMAT` and if the user need to format the time he always need the parsed string.

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @property
+    def description(self):
+        """ pull request description """
+        return self.get_data("description")
+
+    @property
+    def is_declined(self):
+        """ True if the pull request was declined """
+        return self.get_data("state").upper() == PR_STATE_DECLINED
```
Spacetown, line 87: The `upper()` isn't needed because the constants are documented in the API with a exact case. This MUST match.

atlassian/bitbucket/server/__init__.py
```diff
@@ -15,6 +15,7 @@ def __init__(self, url, *args, **kwargs):
         kwargs["cloud"] = False
         kwargs["api_root"] = "rest/api"
         kwargs["api_version"] = "1.0"
+        kwargs["timeformat_func"] = None
```
Spacetown, line 18: Init value shall be the same as in `Cloud`. This line can be deleted.

atlassian/bitbucket/cloud/__init__.py
```diff
@@ -3,13 +3,16 @@
...
+from .const import CONF_TIMEFORMAT
+from datetime import datetime
 
 
 class Cloud(BitbucketCloudBase):
     def __init__(self, url, *args, **kwargs):
         kwargs["cloud"] = True
         kwargs["api_root"] = None
         kwargs["api_version"] = "2.0"
+        kwargs["timeformat_func"] = kwargs.pop("timeformat_func", lambda ts: datetime.strptime(ts, CONF_TIMEFORMAT))
```
Spacetown, line 15: This block shall be moved to `atlassian/bitbucket/base.py` because the attribute is stored there and the server shall use the same value.
```python
kwargs["timeformat_func"] = kwargs.pop("timeformat_func", lambda x: x)
```

atlassian/bitbucket/cloud/base.py
```diff
@@ -85,3 +97,10 @@ def get_data(self, id, default=None):
 
     def get_link(self, link):
         return self.__data["links"][link]["href"]
+
+    def get_time(self, id):
+        value_str = self.get_data(id)
+        if not value_str or not isinstance(value_str, str):
```
Spacetown, line 103: IF the value is `False` the test is true and None is returned.
```suggestion
        if value_str is None or not isinstance(value_str, str):
```

Spacetown (changes requested):

atlassian/bitbucket/cloud/base.py
sti0: But if it's `False` than it's `bool` and the second condition will trigger...this implementation will even solve the falsy string problem when the `string == ""`

Spacetown: When is a time string empty? I think in this case the exception from parsing the time is more suitable. Also if there is a empty string and the user configured raw values, this empty string should be returned.

atlassian/bitbucket/cloud/repositories/defaultReviewers.py
sti0: Yes. Because the user is deleted, the returned object should be empty. But what about `if "errors"...` , IMO this should raise an exception?

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,248 @@
...
+    PRP_CHANGES_REQUESTED,
+    PRP_ROLE_PARTICIPANT,
+    PRP_ROLE_REVIEWER,
+    PR_MERGE_COMMIT,
+    PR_MERGE_STRATEGIES,
+    PR_STATE_DECLINED,
+    PR_STATE_MERGED,
+    PR_STATE_OPEN,
+    PR_STATE_SUPERSEDED,
+)
```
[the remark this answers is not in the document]
sti0: For your sake, removed `const.py`

atlassian/bitbucket/cloud/repositories/defaultReviewers.py
Spacetown: This if block is copied from the old non OO implementation. If the user activates the raw responses a error response will not trigger an exception. In the OO interface the raw response shall never be activated by the user and therefore a exception is thrown. Maybe it can be removed.

sti0: Ok. I think that's no task from this implementation. Therefore it should be clarified / modified in a extra pull request.

atlassian/bitbucket/cloud/repositories/pipelines.py
```diff
@@ -156,11 +156,11 @@ def run_number(self):
 
     @property
     def started_on(self):
-        return self.get_data("started_on")
+        return self.get_time("started_on")
 
     @property
     def completed_on(self):
-        return self.get_data("completed_on")
+        return self.get_time("completed_on")
```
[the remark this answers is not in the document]
sti0: Don't know what you mean. One `completed_on` is for the whole pipeline the other is for the single step.

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,248 @@
...
+        :param merge_strategy: string:  Merge strategy (one of "merge_commit", "squash", "fast_forward")
+        :param close_source_branch: boolean: Close the source branch after merge, default PR option
+        """
+        self._check_if_open()
+
+        if merge_strategy not in PR_MERGE_STRATEGIES:
+            raise ValueError("merge_stragegy must be {}".format(PR_MERGE_STRATEGIES))
+
+        data = {
+            "close_source_branch": close_source_branch or self.close_source_branch,
```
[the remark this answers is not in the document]
sti0: it is:
https://github.com/atlassian-api/atlassian-python-api/blob/6ce4e80a1f32bb3db9425b63f38e8bb1e2351501/atlassian/bitbucket/cloud/repositories/pullRequests.py#L115

atlassian/bitbucket/cloud/repositories/pipelines.py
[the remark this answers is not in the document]
Spacetown: You're right. I didn't saw the blue ruler in the diff view.

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,248 @@
...
+        """ Unapprove a pull request if open """
+        self._check_if_open()
+        return self.delete("approve")
+
+    def decline(self):
+        """ Decline a pull request """
+        self._check_if_open()
+        return self.post("decline")
+
+    def merge(self, merge_strategy=PR_MERGE_COMMIT, close_source_branch=None):
```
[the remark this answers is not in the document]
sti0: Now using `None` as default value. Todo: Adding default project merge strategy if `None` and remove the exception

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: Please check the new implementation. I added a default lambda function which handles the string checks. But you are able now to pass in another lambda function to e.g. return a default value if `None`. Something like this:

```
from atlassian.bitbucket.cloud import Cloud
from atlassian.bitbucket.base import CONF_TIMEFORMAT
from datetime import datetime
import pendulum


def timeformat(x):
    tz = pendulum.timezone("Europe/Berlin")
    if isinstance(x, str):
        x = datetime.strptime(x, CONF_TIMEFORMAT)
        return tz.convert(x)
    else:
        return "no date"


bb = Cloud(
    url="https://api.bitbucket.org",
    username="xxx",       
    password="xxxx",
    timeformat_lambda=lambda x: timeformat(x),
    # timeformat_lambda=None,
)
```

But you still have to import the constant. But I think this implementation is more flexibel for the user. Passing the date convert directly into the lambda would not work with that approach because `strptime` expects a `str`.

Another approach would be:
```
 def get_time(self, id):
        value_str = self.get_data(id)
        if isinstance(value_str, str):
            value = datetime.strptime(value_str, CONF_TIMEFORMAT)
        else:
            value = value_str

        return value_str if self.timeformat_lambda is None else self.timeformat_lambda(value)
```
This would pass the converted `datetime` object or the pure type (if not `str`). Everything else must be handled in the lambda function. In that scenario, the import of `CONF_TIMEFORMAT` is not necessary.

atlassian/bitbucket/cloud/repositories/defaultReviewers.py
```diff
@@ -70,25 +71,9 @@ def get(self, user):
...
-        return self.get_data("nickname")
-
-    @property
-    def account_id(self):
-        return self.get_data("account_id")
-
-    @property
-    def uuid(self):
-        return self.get_data("uuid")
+        super(DefaultReviewer, self).__init__(url, data, *args, **kwargs)
```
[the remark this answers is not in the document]
sti0: This is not necessary because `DefaultReviewer` inherits from `User`

Spacetown:
One general suggestion is it possible to add a link to the API endpoint for the new functions? I've done this in the Server implementation and this was very helpful for me.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
Spacetown: I think the second approach is Vetter because the User doesn't need to take Cafe about the interna format. Getting the local time is quite easy.
```python
def timeformat(x):
    if x:
        return pendulum.timezone("Europe/Berlin").convert(x)
    else:
        return "no date"
```
Converting the time into an object in `get_time` is only needed if a lambda is defined. If the User requests raw strings the conversion isn't needed.
```python
 def get_time(self, id):
        value_str = self.get_data(id)
        if self.timeformat_lambda is None:
            return value_str

        if isinstance(value_str, str):
            value = datetime.strptime(value_str, CONF_TIMEFORMAT)
        else:
            value = value_str

        return self.timeformat_lambda(value)
```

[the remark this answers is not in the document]
sti0: Added this approach.

sti0:
API doc links are added.

atlassian/bitbucket/base.py
```diff
@@ -1,8 +1,12 @@
 # coding=utf-8
 
+from datetime import datetime
 from ..rest_client import AtlassianRestAPI
 
 
+CONF_TIMEFORMAT = "%Y-%m-%dT%H:%M:%S.%f%z"
+
+
```
Spacetown, line 9: Move as attribute of `BitbucketBase`, than it can be accessed with every derived class or instance without an additional import.

atlassian/bitbucket/base.py
```diff
@@ -1,8 +1,12 @@
 # coding=utf-8
 
+from datetime import datetime
```
Spacetown, line 3: I think `datetime` should be added as `install_requires` to `setup.py` and to `requirements.txt`.

atlassian/bitbucket/cloud/repositories/pullRequests.py
```diff
@@ -0,0 +1,312 @@
...
+
+        API docs: https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pullrequests/%7Bpull_request_id%7D/decline
+        """
+        self._check_if_open()
+        return self.post("decline")
+
+    def merge(self, merge_strategy=None, close_source_branch=None):
+        """
+        Merges the pull request if it's open
+        :param merge_strategy: string:  Merge strategy (one of "merge_commit", "squash", "fast_forward")
```
Spacetown, line 254: Better reference the constants in the documentation, e.g. `self.MERGE_COMMIT`

atlassian/bitbucket/base.py
sti0: `datetime` is a Python basic lib. https://docs.python.org/3/library/datetime.html

atlassian/bitbucket/cloud/repositories/pullRequests.py
sti0: done.

atlassian/bitbucket/base.py
sti0: Done.

atlassian/bitbucket/cloud/repositories/pullRequests.py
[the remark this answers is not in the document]
sti0: Tested: If `None`, the project default merge strategy will be used.

sti0:
PR #661 will fix the unit test failures...

atlassian/bitbucket/cloud/repositories/users.py
sti0: This should be implemented in the next iteration. #662

sti0:
This is ready to go.

PR #658, opened 2020-12-23
Add coverage report.

### LINKED ISSUE #657: [Test] Adding coverage report
What about adding coverage report to the test?
New or changed endpoints should also be full covered by the test.

Spacetown:
There was a problem with the different python versions. Effectivly only version 3.8.5, which is the default in the docker ubuntu, was used. Now 3.6, 3.7, 3.8 and 3.9 are used.

At the moment the result is only visible inside the log of tox. To make the result vissible in the PR a project in Codecov is needed and the token must be added to this project to be able to upload the data from the GitHub action. We should also define from which python version the result is uploaded.
You can see in https://github.com/gcovr/gcovr/pull/445 how it can look like.

Should we install all python versions inside the docker container and run tox with all python versions?
Closes #657

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.4', '3.5', '3.6', '3.7', '3.8', '3.9', '-dev']
+        python-version: ['3.6', '3.7', '3.8', '3.9']
 
     steps:
     - uses: actions/checkout@v2
--- a/Dockerfile.qa
+++ b/Dockerfile.qa
@@ -2,20 +2,30 @@ FROM ubuntu:20.04
 
 ARG PYTHON_VERSION
 
-RUN \
-  apt update && \
-  apt-get install -y make gcc-8 libkrb5-dev python$PYTHON_VERSION python3-pip && \
-  apt-get clean
+# Install add-apt-repository
+RUN apt-get update
+RUN apt-get install -y software-properties-common
 
+# Install python development
+RUN add-apt-repository ppa:deadsnakes/ppa
+RUN apt-get update
+RUN apt-get install -y python$PYTHON_VERSION-dev python3-pip
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python$PYTHON_VERSION 1
+
+# Install kerberos development
+RUN apt-get install -y libkrb5-dev
+
+# Clean
+RUN apt-get clean
+
+# Install modules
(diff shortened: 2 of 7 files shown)


Spacetown:
@sti0 I'm fixing the usage of the correct python version in the tests. The test for python 3.4 fails parsing the time format. According to https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes, note 6 the %z was changed in version 3.7 to accept the double colon.
For python 3.4 the double colon should be removed before parsing the string. Should we also do this always?

sti0:
@Spacetown Do you mean this? https://github.com/atlassian-api/atlassian-python-api/blob/61affb06653e0b4ba564a6e5ef0c60ff63a9c42c/tests/test_bitbucket_cloud.py#L26

Spacetown:
> @Spacetown Do you mean this?
> 
> https://github.com/atlassian-api/atlassian-python-api/blob/61affb06653e0b4ba564a6e5ef0c60ff63a9c42c/tests/test_bitbucket_cloud.py#L26

Sorry, I forgot the link to the log: https://github.com/atlassian-api/atlassian-python-api/pull/658/checks?check_run_id=1627624974#step:3:1824

The problem is
https://github.com/atlassian-api/atlassian-python-api/blob/61affb06653e0b4ba564a6e5ef0c60ff63a9c42c/atlassian/bitbucket/base.py#L80

sti0:
Ok, now I know what you mean. Think we could replace the double colon for the timezone.

Spacetown:
> Ok, now I know what you mean. Think we could replace the double colon for the timezone.

Only for version <= 3.6 or for all versions? I prefer <= 3.6.

sti0:
> > Ok, now I know what you mean. Think we could replace the double colon for the timezone.
> 
> Only for version <= 3.6 or for all versions? I prefer <= 3.6.

To prevent sideeffects, yes.

Spacetown:
@gonchik Now the code is different between the minor versions. Therefore we should keep testing all available minor versions.

sti0 (approved):

Spacetown:
@gonchik Shall the upload of the report to Codecov be added in a seperate PR? This can only be done by a project admin because a project on Codecov must be created and linked to Github.

gonchik:
@Spacetown  yeap, let's do in separate PR.

Spacetown:
@gonchik Than this PR is finished. Python 3.4 and 3.5 have to be removed because there is no dev version available inside docker. The dev version is needed to compile kerberos.

gonchik:
Got it! :) Thank you!

PR #661, opened 2020-12-27
[Bitbucket Cloud] Test-Mockup encoding / exec global

sti0:
This forces the mockup encoding to `utf-8`.  Rest client expects a `utf-8` encoding:
https://github.com/atlassian-api/atlassian-python-api/blob/5028c00622673e36fa447846b8c3bc05ad0143de/atlassian/rest_client.py#L224

Prior Python 3.9: without `encoding` files are read with system default codepage, e.g. windows  = `cp1252`.

E.g. [Python 3.8](https://docs.python.org/3.8/library/functions.html#open)

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever locale.getpreferredencoding() returns), but any text encoding supported by Python can be used. See the codecs module for the list of supported encodings.

Additionaly the exec global `"null": None` is added, because BB cloud api may return the value `null` instead of `None`.


Spacetown:
For which response is the `"null"` needed?

sti0:
I simply dumped the json from the pull request response for unit tests and some attributes are `null`. 
At the docs they can be `null`:
image

Spacetown:
Ok, in my previous dumps there wasn't a `null`.

Spacetown (approved):
LGFM

gonchik:
Thank you!

PR #666, opened 2020-12-30
Auth: Add OAuth 2.0 authentication support

jensvog:
The current implementation only supports authentication with OAuth 1.0. They are mostly used by the server variants. The cloud services now use OAuth 2.0.

With this commit OAuth 2.0 is now supported without breaking the support for OAuth 1.0. I also extended the documentation and created an example. The example shows a server based on Flask. It does the handshake with Bitbucket and receives the tokens. I also deployed once and tested it successfully.

This can be helpful for some web apps that need to interact with Atlassian services.

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -4,7 +4,7 @@ from json import dumps
 
 import requests
 from oauthlib.oauth1 import SIGNATURE_RSA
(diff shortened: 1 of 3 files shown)


gonchik:
Awesome! Thank you!

PR #671, opened 2021-01-01
Docs: Get documentation version from VERSION file

jensvog:
The version of the documentation is configured in the `conf.py` file. The version has not been increased there for a longer time. On Read the Docs you can find the version information for the library in the title for example. So it looks like the documentation refers to an older version of the library.
version_documentation

The current version of the library is in the `atlassian/VERSION` file. In oder to keep the process simple, I changed the `conf.py` so that the documentation now also takes the version information from the `VERSION` file. By that the version information of the library and the documentation is always in sync.

I set both `release` and `version` to the same value. I think there is no distinction needed for this project.

I tested it locally and it worked well.

[Background informations](https://www.sphinx-doc.org/en/master/usage/configuration.html)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,7 +19,11 @@ author = "SLRover"
 # The short X.Y version
 version = ""
 # The full version, including alpha/beta/rc tags
-release = "1.15.3"
+release = ""
+
+with open(os.path.join("../atlassian", "VERSION")) as file:
+    version = file.read().strip()
+    release = version
 
 # -- General configuration ---------------------------------------------------
 


docs/conf.py
```diff
@@ -19,7 +19,11 @@
 # The short X.Y version
 version = ""
 # The full version, including alpha/beta/rc tags
-release = "1.15.3"
+release = ""
+
+with open(os.path.join("../atlassian", "VERSION")) as file:
```
Spacetown, line 24: This relies on the working directory of the shell. To be independant from the working directory the variable __file__ should be used. But at the moment there are more lines which relies on the working directory.

Spacetown (approved):

Spacetown:
@gonchik You should remove python 3.4 and 3.5 as required tests because they aren't available anymore.

gonchik:
Curious I can't find anywhere the dependency of 3.4. 
I remember it has been removed from test.yml file

PR #672, opened 2021-01-01
[Bitbucket] Unify Server and Cloud implementation

Spacetown:
Unify the OO implementations:
- Remove checks for "error" in response because an exception is thrown on HTTP errors.
- Moved `delete` from base class because of the API link. Always returns nothing.
- Moved `update` from base class because of the API link. Always returns the updated objects (`self`).
- Add missing link to API documentation always at the end and move existing links. Links always start with "API docs:".
- Fixed relative links in Cloud pull request test mookup and use this links instead of building own links in code.
- Fix update of properties for Server.


gonchik:
Thank you! :)

PR #673, opened 2021-01-02
All: Fixes typos

jensvog:
Scanned over whole repository and fixes some typos.

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -1426,7 +1426,7 @@ class Bitbucket(BitbucketBase):
         data = {"version": comment_version, "text": comment}
         return self.put(url, data=data)
 
-    def delete_pull_reques_comment(
+    def delete_pull_request_comment(
         self,
         project_key,
         repository_slug,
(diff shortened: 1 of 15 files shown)


Spacetown:
Fixes in comments are welcome but fixes in function names are incompatible changes.

gonchik:
@Spacetown  as I see only 2 method were renamed. 

I think better to merge with small wrapper rule. 

 ``` 
@deprecated 
def delete_pull_reques_comment(self,repository_slug,
        pull_request_id,
        comment_id,	
        comment_version,	        
    ):
"""
Typo will be removed
"""
 return self.delete_pull_request_comment(repository_slug,
        pull_request_id,
        comment_id,	
        comment_version)
```

Spacetown:
The second one already had a wrapper rule but with a log entry instead of `@deprecated` decorator.

jensvog:
Thanks for your feedback. I created an update.

* The confluence method already had a hint and a corrected method. Thanks for the hint @Spacetown. I reverted the typo change there.
* I updated the bitbucket function so that the typo method is marked as deprecated and added a method with correct function name, like you supposed @gonchik.

How does it look now to you?

Issue #638, opened 2020-11-28
Confluence: Allow get_group_members() to page through results


bobgott:
Confluence.get_group_members() returns response['results'] which works correctly if only 1 request is required.  In my case where there are many members per group, I need to page through the results by incrementing 'start' accordingly.

Perhaps you might want to return just 'response'?

This is easy to work around but thought you might want to address it in your code as well.

[feature request]

gonchik:
Hi! 

@bobgott  Sorry for the delayed answer. 
So you can iterate until you did not get empty collection. 
In that commit provided the more clear example and added the method to provide all users
https://github.com/atlassian-api/atlassian-python-api/commit/1067b58288b2522ce6a45feecd3695e6cb99e991

gonchik:
examples/confluence/confluence_get_group_members.py

PR #674, opened 2021-01-03
[Confluence] Get group members

gonchik:
#638 provide a method to reduce questions related to fetch all users


atlassian/confluence.py
```diff
@@ -1737,6 +1737,29 @@ def get_group_members(self, group_name="confluence-users", start=0, limit=1000,
...
+        :param group_name
+        :param expand: OPTIONAL: A comma separated list of properties to expand on the content. status
+        :return:
+        """
+        limit = 50
+        flag = True
+        step = 0
+        members = []
+        while flag:
+            values = self.get_group_members(group_name=group_name, start=limit * step, limit=limit, expand=expand)
```
Spacetown, line 1752: What if server limit is below limit?
```suggestion
            values = self.get_group_members(group_name=group_name, start=len(members), limit=limit, expand=expand)
```

atlassian/confluence.py
```diff
@@ -1737,6 +1737,29 @@ def get_group_members(self, group_name="confluence-users", start=0, limit=1000,
...
+            step += 1
+            if len(values) == 0:
+                flag = False
+            else:
+                for value in values:
+                    members.append(value)
```
Spacetown, line 1758:
```suggestion
                members.extend(values)
```

PR #675, opened 2021-01-03
Fix makefile

Spacetown:
This wasn't a argument, it was the build script.


gonchik:
Thanks for validating :)

PR #676, opened 2021-01-04
Docs: Ignore build folder in doc8

jensvog:
When documentation has been built and quality check is run afterwards, it always fails because of errors in the docs/_build/ folder. This is because doc8 is also scanning the generated documentation files for errors.
doc8_error

This commit is adding an ignore-parameter to the call to doc8. With this the build folder is ignored.


gonchik:
Thank you

PR #682, opened 2021-01-05
Docs: Remove Key-Cert authentication info

jensvog:
Documentation gave information about authentication with key and certificate. I could not find this authentication method in the code and as well not in the Atlassian documentation.

Therefore, I removed the chapter is removed and the other authentication methods are grouped with a new title. Maybe you have more information what could be meant there?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -49,45 +49,10 @@ Add a connection:
         username='admin',
         password='admin')
 
-Key/Cert Based authentication
------------------------------
+Other authentication methods
+----------------------------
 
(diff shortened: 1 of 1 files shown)


Spacetown (approved):
LGFM because the options are missing in
https://github.com/atlassian-api/atlassian-python-api/blob/6ffed814288d09a755f04c59108d66f9f7a8e59d/atlassian/rest_client.py#L37-L54

gonchik:
Thank you! :)

PR #683, opened 2021-01-05
[JIRA] add example jira_add_label_tag.py

copolycube:
thanks for the discovery of the "black" python formatting tool !

### DIFF
--- a/examples/jira/jira_add_label_tag.py
+++ b/examples/jira/jira_add_label_tag.py
@@ -1,0 +1,24 @@
+# coding=utf-8
+from atlassian import Jira
+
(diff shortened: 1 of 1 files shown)


Spacetown:
@copolycube Joke or serious?

copolycube:
@Spacetown , no seriously I didn't know that tool.

For the pull request, it's just a basic example I tested since I wasn't sure the Labels (system field) would actually be modifiable the same way a _custom_ field is, thought it could help some other people too, but feel free to reject if it's against your policy.

gonchik:
Thank you! :)

Issue #681, opened 2021-01-05
[Confluence] post method missing advanced mode parameter


VTimofeenko:
Hi!

I want to include result of POSTing to a private Confluence API in my code to convert markdown to storage format. The way I am trying to use it is:

```python
    url = "rest/tinymce/1/markdownxhtmlconverter"
    # the endpoint returns plain text, need to redefine the default header
    headers = {"Content-Type": "application/json"}
    response = confluence.post(url, data={"wiki": text}, headers=headers)
    return response
```

The output will include a warning and is only intended to be used as fallback.

This endpoint returns plaintext, rather than json:

```
> curl -XPOST -u 'admin:pwd' http://localhost:8090/rest/tinymce/1/markdownxhtmlconverter -H 'Content-Type: application/json' -d '{"wiki": "# AAA\n * one\n * two"}'
<h1>AAA</h1> <ul> <li>one <li>two </ul>
```

so [`_response_handler`](https://github.com/atlassian-api/atlassian-python-api/blob/93a47b116ff7fdf4bd092f3780f9dff05d1ff67b/atlassian/rest_client.py#L139) cannot convert it.

In the [get method definition](https://github.com/atlassian-api/atlassian-python-api/blob/93a47b116ff7fdf4bd092f3780f9dff05d1ff67b/atlassian/rest_client.py#L250), it has a separate parameter "[`advanced_mode`](https://github.com/atlassian-api/atlassian-python-api/blob/93a47b116ff7fdf4bd092f3780f9dff05d1ff67b/atlassian/rest_client.py#L260)", but [post](https://github.com/atlassian-api/atlassian-python-api/blob/93a47b116ff7fdf4bd092f3780f9dff05d1ff67b/atlassian/rest_client.py#L298) method does not have it, relying on advanced_mode property of the AtlassianRestAPI instance. Is this by design?

If not, I would be glad to whip up a quick PR.

Spacetown:
The `advanced_mode` was only needed for the GET. You can add it to the POST or better to all other request methods.

VTimofeenko:
@Spacetown , thanks! I will do that. Does it have other intended meaning other than making the underlying methods return raw response?

Spacetown:
As you can see, also the error handling is turned of.
https://github.com/atlassian-api/atlassian-python-api/blob/6ffed814288d09a755f04c59108d66f9f7a8e59d/atlassian/rest_client.py#L229-L248

PR #684, opened 2021-01-05
[RestClient] Added advanced_mode to POST, PUT, DELETE

VTimofeenko:
Added boolean "advanced_mode" switch to post, put and delete methods of RestClient, to make them align with the "[get](https://github.com/atlassian-api/atlassian-python-api/blob/93a47b116ff7fdf4bd092f3780f9dff05d1ff67b/atlassian/rest_client.py#L250)" implementation. This allows the caller to retrieve raw response for individual calls rather than change the property of the RestClient instance.

Added tests that mimic Confluence module behavior and check the type of response.

Closes atlassian-api/atlassian-python-api#681

### DIFF
--- a/tests/test_confluence_advanced_mode.py
+++ b/tests/test_confluence_advanced_mode.py
@@ -1,7 +1,6 @@
 # coding=utf-8
 import json
 import os
-import tempfile
 import unittest
 from requests import Response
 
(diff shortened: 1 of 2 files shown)


tests/test_confluence_advanced_mode.py
```diff
@@ -0,0 +1,114 @@
...
+                credentials = json.load(json_file)
+                cls.confluence = Confluence(
+                    url=credentials["host"],
+                    username=credentials["username"],
+                    password=credentials["password"],
+                )
+        except FileNotFoundError:
+            raise unittest.SkipTest("credentials.secret missing, skipping test")
+        except Exception as err:
+            self.fail("[{0}]: {1}".format(self.secret_file, err))
```
Spacetown, line 43: Is the surrounding `try` realy needed? If the file can't be read or the connection fails a exception is raised and the test fails.

VTimofeenko: The FileNotFound explicit handling is probably not needed, since it duplicates the skipIf condition for the class.

I left the Exception handling to keep the test similar to the other ones.

Spacetown (approved):
LGFM

gonchik:
Thank you!

PR #685, opened 2021-01-06
Bitbucket Cloud: Set default url

jensvog:
Unlike the server variants, the url for the Bitbucket cloud is always the same. There is no need to write the url every time, when the Cloud class should be used. Also for new users of this library it can be confusing, where to find the Bitbucket cloud url.

Therefore, this commit sets for the url parameter of the Bitbucket Cloud class to a default value. It still can be modified by giving another url as parameter (i.e. for tests). Also adapted documentation and examples.

### DIFF
--- a/atlassian/bitbucket/cloud/__init__.py
+++ b/atlassian/bitbucket/cloud/__init__.py
@@ -6,7 +6,7 @@ from .repositories import Repositories
 
 
 class Cloud(BitbucketCloudBase):
-    def __init__(self, url, *args, **kwargs):
+    def __init__(self, url="https://api.bitbucket.org/", *args, **kwargs):
         kwargs["cloud"] = True
         kwargs["api_root"] = None
         kwargs["api_version"] = "2.0"
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -121,7 +121,7 @@ OAuth 2.0 is also supported:
 
 .. code-block:: python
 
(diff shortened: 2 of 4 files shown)


Spacetown:
I do not agree with this because many companies have a private cloud instance with another url. Someone who wants to use the API should know the root url.

jensvog:
Maybe I have a misunderstanding, but I thought for self-hosting you have the server variants, which you can trigger with the atlassian.bitbucket.Server. But the cloud is the one hosted by Bitbucket itself.

gonchik:
@jensvog  Thank you for that idea, that's really good. 
Can we keep url = "api.bitbucket.org" for examples ? 
Because it will help to users to understand the url goes to api instance.

Spacetown:
@jensvog The Server family is discontinued and you can switch to data center. I thought this is something between.

Spacetown:
In the documentation the `.cloud` should be removed from the import statement.

jensvog:
Thanks for the feedback. Ok I updated the PR. How does it look now for you? Should it also be reverted in the documentation, or is examples enough?

Spacetown:
@jensvog As mentioned, in the example the Import was changed, this should also be done in `Index.rst` line 124 and 227.

jensvog:
@Spacetown Thanks, did it

Spacetown (approved):
LGFM

gonchik:
Thank you! :)

PR #691, opened 2021-01-19
[Jira Service management] Update service_desk.py

spocki:
with the default 'application/json' content-type the upload is not working
in the documentation there is a hint which headers should be set
if the X-Atlassian-Token is missing you receive a 404


gonchik:
Thanks for the PR.  @spocki 
Could you wrap into that place the headers information?
https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/rest_client.py#L31
I think good to have in one place a header information, because in the future it will be so hard to support in the future. 

If you have other opinion, feel free to provide

gonchik:
Thank you!

PR #692, opened 2021-01-19
[REST Client] Kerberos authentication improvements

offa:
The package currently used on windows is broken on Python 3.9 (PR pending, but the project isn't very active). This PR replaces both kerberos packages with [`requests-kerberos`](https://github.com/requests/requests-kerberos), allowing some improvements:

- Single package, no difference for windows
- API is closer to *oauth*, using a map (the parameter isn't used yet)
- Less code
- Python 3.9 compatibility

##### Why not _requests-gssapi_?

It requires external installation of a additional executables on Windows.

## TODO / Discussion

- [x] Docs: Update example
- [x] Is `OPTIONAL` a good default for mutual authentication (if not, it should be configurable though)?
- [x] What parameters should be configurable through the parameter?

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -87,21 +87,10 @@ class AtlassianRestAPI(object):
     def _create_basic_session(self, username, password):
         self._session.auth = (username, password)
 
-    def _create_kerberos_session(self, kerberos_service):
-        try:
-            import kerberos as kerb
-        except ImportError as e:
-            log.debug(e)
-            try:
-                import kerberos_sspi as kerb
-            except ImportError:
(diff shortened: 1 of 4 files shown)


gonchik:
@offa  regarding kerberos I don't use it, suddenly. 

> Single package, no difference for windows
Fully support you

> API is closer to oauth, using a map (the parameter isn't used yet)
+1

offa:
Since I don't have a suitable parameter to pass yet, shall I update the example using an empty map instead?

gonchik:
Yeap, please do it

offa:
I have updated the example.

gonchik:
Thank you!

PR #694, opened 2021-01-19
[Test] Upload coverage data to codecov.io.

Spacetown:
@gonchik For public repositories no token is needed. Th codecov.io can be marked as needed request. It fails if the converage decrease.


gonchik:
That's great  :)

Issue #696, opened 2021-01-20
confluence.py attach_content uses depreciated value for "X-Atlassian-Token" field HTTP header


pscheir:
confluence.py attach_content function uses a depreciated HTTP header value for the "X-Atlassian-Token" field

"X-Atlassian-Token" should be set to “no-check” instead of “nocheck”.
See https://developer.atlassian.com/server/confluence/form-token-handling/

Otherwise the following warning will appear in the server log: 
[common.security.jersey.XsrfResourceFilter] hasDeprecatedHeaderValue Use of the 'nocheck' value for X-Atlassian-Token has been deprecated since rest 3.0.0. Please use a value of 'no-check' instead.

Also see PR #697

gonchik:
Thanks for the fix #697

PR #697, opened 2021-01-20
Correct value for "X-Atlassian-Token" field in HTTP header - confluence.py (attach_content)

pscheir:
X-Atlassian-Token should be set to “no-check” instead of “nocheck”.
See https://developer.atlassian.com/server/confluence/form-token-handling/

Otherwise the following warning appears in the server log:
[common.security.jersey.XsrfResourceFilter] hasDeprecatedHeaderValue Use of the 'nocheck' value for X-Atlassian-Token has been deprecated since rest 3.0.0. Please use a value of 'no-check' instead.

Also see issue #696

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -846,7 +846,7 @@ class Confluence(AtlassianRestAPI):
                 "comment": comment,
                 "minorEdit": "true",
             }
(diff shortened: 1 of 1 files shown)


gonchik:
Thank you!

Issue #704, opened 2021-01-28
Regression in 3.3.0


hograthm:
Commit a20bb60b introduced regression for Python 2.7 users:
https://github.com/atlassian-api/atlassian-python-api/blob/c6673d8ba904fff944cbc5398dafb6c6d5a3eea1/atlassian/bitbucket/cloud/repositories/pullRequests.py#L366
PEP 3107 Function Annotations `->` is not supported in Python 2.7
More about function annotations here:
https://python-future.org/func_annotations.html

PR #705, opened 2021-01-28
[Bitbucket] Fix Python 2.7 incompatibility

hograthm:
Remove function annotation since this syntax is not supported in Python 2.7
Fix for #704


hograthm:
FIx for #704

PR #714, opened 2021-02-01
Update __init__.py

manefix:
fix issue with incorrect api call for declining PRs


gonchik:
Thanks

manefix:
@gonchik could you please share schedule for regular python wheel creation to get this change on all my computers? Is there any chance to rebuild new wheel manually?

PR #721, opened 2021-02-14
Enhanced set_issue_status method in jira module, adding update param

shyammeduri:
We have a specific requirement in our work area, where for some issue status changes, we are required to mandatorily provide comments. The current implementation of set_issue_status method doesn't support updating comments, as comments can't be updated via 'fields' and instead should be updated via the 'update' json structure as per the latest Jira Rest API. Instead of creating a bug/enhancement request I thought I can fix it as it is a simple fix. I have tested this fix and it works fine.

### DIFF
--- a/atlassian/bamboo.py
+++ b/atlassian/bamboo.py
@@ -357,7 +357,7 @@ class Bamboo(AtlassianRestAPI):
             elements_key="results",
             element_key="result",
             label=label,
-            **params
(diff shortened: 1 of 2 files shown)


gonchik:
Hi @shyammeduri , 
Please, check via black the formatting

shyammeduri:
Formatted with black

Issue #722, opened 2021-02-15
Drop f-string from package


StykMartin:
Hello,

I tried to use this package with Python 3.4 and it doesn't work.
The reason for that is it contains f-strings. This was introduced in Python 3.6.

When I checked the supported version, it seems that even Python 2 is supported at this moment.

PR #723, opened 2021-02-15
Use format instead of f-strings

StykMartin:
We are still running the atlassian-python-api package on Python 3.4, due to < reasons/enterprise world>.
When I check, this package should be supported even with Python 2.7, therefore, f-string shouldn't be used.

Ref https://github.com/atlassian-api/atlassian-python-api/blob/master/setup.py#L45

I removed the f-string only from package sources themselves. 
I didn't touch tests nor examples.

Fixes: #722.

Signed-off-by: Martin Styk <mart.styk@gmail.com>


StykMartin:
@gonchik opinion?

gonchik:
Thank you!

PR #727, opened 2021-02-24
[JIRA] User search required parameters additions

### LINKED ISSUE #726: [Jira] Jira.user_find_by_user_string() method does not support required parameters
According to the [documentation](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-user-search/#api-rest-api-2-user-search-get) the Jira user search endpoint requires specifying an `accountId`, `property` or `query` parameter. Currently, we don't support any of those parameters and we require a `username` parameter, which is optional in the REST API and in strict GDPR cases it is an invalid parameter which results in an HTTP 400 error when provided.

This should mean the `Jira.user_find_by_user_string()` method is essentially broken for everyone as we cannot pass any of the required parameters.

According to the REST API documentation, if an `accountId` parameter is passed, it cannot be passed along with a `query` parameter or else it will return an HTTP 400 error. It's not clear in the API documentation whether the API will accept an `accountId` along with a `property` parameter, this should be tested and clarified before implementing these new parameters.

dephekt:
The GDPR changes to the REST API have removed the username parameter from the user search endpoint and now requires either a query, accountId or property parameter be specified otherwise the request will fail. According to the API documentation, a query and accountId cannot both be provided otherwise the request will fail.

This PR brings the Python Jira.user_find_by_user_string() method into alignment with the REST API's post-GDPR changes. It also updates the associated documentation per the contribution guidelines. I tested the functionality against Jira Cloud that my company uses and it worked as expected.

Fixes #726. See that issue for more details. This method is broken until this PR is merged.

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -1472,21 +1472,21 @@ class Jira(AtlassianRestAPI):
                 params["accountId"] = account_id
             elif account_id and query:
                 return "You cannot specify both the query and account_id parameters"
-            elif not account_id and not query and not property_key:
+            elif not any([account_id, query, property_key]):
                 return "You must specify at least one parameter: query or account_id or property_key"
             elif username:
                 return "Jira Cloud no longer supports a username parameter, use account_id, query or property_key"
+
+            if query:
+                params["query"] = query
+            if property_key:
+                params["property"] = property_key
         elif not username:
             return "Username parameter is required for user search on Jira Server"
-
-        if username:
+        elif any([account_id, query, property_key]):
+            return "Jira Server does not support account_id, query or property_key parameters"
+        else:
             params["username"] = username
-
-        if self.cloud and query:
-            params["query"] = query
-
-        if self.cloud and property_key:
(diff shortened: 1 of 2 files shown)


gonchik:
Hi! 
As I see for on-prem checked on the supported versions till 8.5, still used old way :( .

dephekt:
@gonchik OK, so what I'm doing now instead is checking if `self.cloud` is true. If it is, then apply the logical checks to ensure that:

- `username` was not specified
- `account_id` or `query` were not both specified
- At least one of `account_id`, `query` or `property_key` were specified

Then, assuming we haven't failed yet, add the params for:
- `account_id`, if `query` wasn't specified
- `query`, if `account_id` wasn't specified
- `property_key` if it was specified

Then if `self.cloud` wasn't true, we require that `username` is specified and that none of the 3 Jira Cloud-specific params were specified and, finally, if we haven't returned an error yet, we add the `username` parameter and perform the search.

gonchik:
Thank you! :)

Issue #515, opened 2020-06-06
Assigning issues is broken on Atlasssian Cloud due to name parameter deprecation


stopthatcow:
Referring to users by "name" is no longer allowed due to GDPR and seems to break assigning of issues using assign_issue() (Client 400 error). I believe we need to replace "name" with "accountId" per https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-user-privacy-api-migration-guide/

PR #728, opened 2021-02-24
Bug/fix assign issue

kkjorsvik:
https://github.com/atlassian-api/atlassian-python-api/issues/515#issue-632122274
The above issue points out the assign_issue using username has been deprecated and now requires the accountId instead. I have updated the code to use accountId.


dephekt (commented):
The `assignee` parameter should probably be changed to reflect the name of the new parameter. E.g. change the input parameter from `assignee` to `account_id` to match the style of other methods using an accountId parameter.

It's a little confusing when the parameter doesn't match the REST API parameter. E.g. this probably should have been called `name` rather than `assignee` originally, so now is a good time to fix that too, IMHO.

While you're at it, it's probably a good time to update the `docs/jira.rst` to document the usage of the method? It's currently undocumented but it's a commonly used method.

kkjorsvik:
Thank you, and I agree with you that those are good changes that should be apart of this pull request. I will work on cleaning things up a bit more. Should I change this Pull Request to a draft or will it be ok to leave it open?

dephekt:
Good question. It's probably OK to leave it open but I personally keep mine as drafts until I'm ready for it to be potentially merged, just to avoid any accidents. Then when I think it's ready, I switch it back.

PR #731, opened 2021-02-25
[Confluence] Update template methods as they are no longer experimental

### LINKED ISSUE #715: Template REST API endpoint not `experimental` anymore
https://github.com/atlassian-api/atlassian-python-api/blob/615219e9d4103ec9807015c4a1f30b36671b3bc9/atlassian/confluence.py#L674

The endpoint changed to `<>/wiki/rest/api/template/<>`, perhaps also with other versions.

I guess now that it's not experimental anymore, would be worth adding this feature to the API documentation too.

dephekt:
Confluence template API transitioned to a stable version. Some things in our module are no longer in line with reality. The endpoint URL has changed to `wiki/rest/api/template` and some parameters have changed behavior. When getting templates or blueprints, a space is not required. If it's not given, all templates in the scope will be returned. I updated the parameters to not be required. The limit has also been updated to a default of 25. I switched those limit parameters to have a default "None" type, so if they're not specified, we don't pass them and we'll get the default API behavior.

We don't currently support deleting templates or creating/updating new content templates but there are REST API methods for that, so now is a good time to write that support.

Fixes #715

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -720,7 +720,7 @@ class Confluence(AtlassianRestAPI):
 
         return self.post("wiki/rest/api/template", json=data)
 
-    def get_template_by_id(self, template_id):
+    def get_content_template(self, template_id):
         """
         Get a content template.
 
@@ -746,9 +746,10 @@ class Confluence(AtlassianRestAPI):
 
         return response
 
-    def get_all_blueprints_from_space(self, space=None, start=0, limit=None, expand=None):
+    def get_blueprint_templates(self, space=None, start=0, limit=None, expand=None):
         """
         Gets all templates provided by blueprints.
+
         Use this method to retrieve all global blueprint templates or all blueprint templates in a space.
         :param space: OPTIONAL: The key of the space to be queried for templates. If ``space`` is not
             specified, global blueprint templates will be returned.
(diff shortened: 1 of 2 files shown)


dephekt:
Drafting this until I finish writing the support for the three extra methods we don't support yet: create, update, and delete.

dephekt:
This needs tested before merging. I don't use or have Confluence in my org so if someone could test the new methods and confirm the old methods that I touched here are working that would be awesome :tada:

I left instructions here on what exactly needs checked: https://github.com/atlassian-api/atlassian-python-api/issues/715#issuecomment-785861951

dephekt:
I also want to refactor the method names to be in-line with the upstream API methods, see this comment since I posted it in the issue instead of here by accident: https://github.com/atlassian-api/atlassian-python-api/issues/715#issuecomment-785990108

dephekt:
This is ready to be merged as soon as someone is able to test the template methods.

dephekt:
Just rebasing off the current master. This is still ready to merge.

dephekt:
Need to readd old method names and add depreciation notice.

gonchik:
@dephekt  Thank you! 👍🏽

PR #732, opened 2021-02-25
[JIRA] Fix get_plugins_info method request in need of trailing slash

### LINKED ISSUE #703: [Jira] 404 on get_plugins_info() in jira module
i'm trying to get all installed plugins info on my jira instance, i've tried to use get_plugins_info() 
but i noticed that get_plugins_info() calls self.get(...) without trailing=True 
which leads the method to send a get to [ {jira_url}/jira/rest/plugins/1.0 ] without a trailing slash 
which leads to a 404 HttpError 
i've tried to manually edit the code and add trailing=True to self.get()
and it worked  
i've check this matter on jira 7.13.0 and not quite sure about this endpoint on other versions 
can you please fix this one ?

dephekt:
The `Jira.get_plugins_info()` method is currently broken. Calling this method results in a 404 client error:

```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://platformsh.atlassian.net/rest/plugins/1.0
```

The issue is the URL should have a trailing slash but we don't set `trailing=True` when calling the `Jira.get()` method in the return. This remedies that situation and fixes #703.

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -2383,7 +2383,7 @@ api-group-workflows/#api-rest-api-2-workflow-search-get)
         :return a json of installed plugins
         """
         url = "rest/plugins/1.0/"
-        return self.get(url, headers=self.no_check_headers)
+        return self.get(url, headers=self.no_check_headers, trailing=True)
 
     def upload_plugin(self, plugin_path):
         """


gonchik:
@dephekt  Thank you !

PR #733, opened 2021-02-26
Add Confluence update_page(always_update) and get_space(params) methods

wlupton:
- update_page(always_update=False) allows the content check to be overridded
- get_space(params=None) allows params to be passed, e.g. type and status


gonchik:
Thank you!

PR #743, opened 2021-03-18
Bulk issue  field update

rakesh556575:
Bulk issue  filed update


gonchik:
@rakesh556575  could you run black formatter please ?
It will be good to have a few works in jira.rst doc

PR #749, opened 2021-04-03
Jira, Confluence: Exceptions contain message provided by server

ValHaris:
When the server returns a 4xx or 5xx status code and provides an
error message, that error message will be the message of the exception raised.

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -2638,3 +2638,18 @@ class Confluence(AtlassianRestAPI):
         url = "rest/jira-metadata/1.0/metadata/cache"
         params = {"globalId": global_id}
         return self.delete(url, params=params)
+
+    def raise_for_status(self, response):
+        """
+        Checks the response for an error status and raises an exception with the error message provided by the server
+        :param response:
+        :return:
+        """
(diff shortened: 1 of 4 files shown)


ValHaris:
I'm still throwing an `HTTPError` instead of an `APIError` in order not to change the API.

The Confluence module has a lot of specific error message hard coded, but the ones provided by the server are at least as good as the hard coded ones for the ones I checked. 
We could consider to remove that code, but that would change the Exception types from `APIError`  to ` HTTPError`

gonchik:
@ValHaris What do you think wrap into dedicated class? 
 https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/errors.py

ValHaris:
@gonchik  : I don't really understand your question.
As I wrote before, I considered both `APIError` and `HTTPError` and decided for the latter so the code behaves exactly like before. An application that just catches `HTTPError` , because the old interface was throwing exactly that, would not need to change anything.
I don't know how strictly this is handled in this project here, but a change to the API should in my opinion only be done for a major version change. And exchanging the exceptions thrown is in my opinion an API change.

PR #752, opened 2021-04-09
[JIRA] add parameter `validateQuery` for search API

KevinDeJong-TomTom:
Allows users to set the parameter `validateQuery` for the Search API. This can be useful in situations where a request is made for a list of issue keys, containing a deleted issue. Disabling the validation is unsafe (to be used by clients on their own risk), but will prevent:

```
{"errorMessages":["An issue with key 'DELETED-1' does not exist for field 'key'."],"errors":{}}
```
and returns a list of valid issues instead (w/o the above `DELETED-1`)


gonchik:
Thanks

PR #753, opened 2021-04-10
Correcting Atlassion to Atlassian

btsteve:
Documents have atlassion instead of atlassion.


gonchik:
Thanks for fix the typo

PR #754, opened 2021-04-13
[JIRA] create_agile_board method adapted for Jira Server

tkuehne-deliveryhero:
The `location` parameter of the `create_agile_board` must not be sent
for the Jira Server edition.
This parameter is not supported as per the [REST API documentation](https://docs.atlassian.com/jira-software/REST/8.5.8/#agile/1.0/board-createBoard) and
throws an error if sent anyway.


gonchik:
Thanks!

Issue #740, opened 2021-03-09
[Bitbucket] Cloud Pipelines pagination is not working


durnerj:
Using this minimal setup, i can only get 10 entries from the API (First Page)

```python
from atlassian.bitbucket.cloud import Cloud

BB_USERNAME = "username"
BB_APP_PASSWORD = "secret"  # Atlassian App-Password with read access
BB_REPO_NAME = "repo-name"
BB_WORKSPACE_NAME = "workspace"

def get_builds():
    bitbucket = Cloud(
        username=BB_USERNAME,
        password=BB_APP_PASSWORD,
        )

    repo = bitbucket.workspaces.get(BB_WORKSPACE_NAME).repositories.get(BB_REPO_NAME)
    data = repo.pipelines.each()
    
    builds = {}
    buildnr = 0
    for build in data:
        buildnr += 1
        builds[buildnr] = build
    return builds

if __name__ == "__main__":
    builds = get_builds()
    print(len(builds))
```

durnerj:
Atlassian confirmed it's an API bug on their side. They are not sending "next" in the Pagination context here.

durnerj:
Issue over there at Atlassian:

https://jira.atlassian.com/browse/BCLOUD-20639

Spacetown:
@gonchik I think you can close this.

gonchik:
Thank you!

PR #755, opened 2021-04-13
[BITBUCKET] Allow users to use filter on pipeline

Amnesycks:
Adding filter "page" & "pagelen" on pipeline.each()
This can allow to use workaround on #740 as mention in https://jira.atlassian.com/browse/BCLOUD-13806

### DIFF
--- a/atlassian/bitbucket/cloud/repositories/pipelines.py
+++ b/atlassian/bitbucket/cloud/repositories/pipelines.py
@@ -59,7 +59,7 @@ class Pipelines(BitbucketCloudBase):
 
         return self.__get_object(self.post(None, trailing=True, data=data))
 
-    def each(self, q=None, sort=None):
+    def each(self, q=None, sort=None, pagelen=None, page=None):
         """
         Returns the list of pipelines in this repository.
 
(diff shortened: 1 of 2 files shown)


Spacetown:
@gonchik I think it is better to do the pagination workaround inernal and not add parameters for this. Also the documentation for the parameters is missing.

Amnesycks:
The pagination on API, work with GET parameters like "sorting", it's why I added this with parameters. About documenting I don't see where the parameters are, only methods. Tell me where and I will update this soon when I got a bit of time to work on it.

Spacetown:
@Amnesycks The arguments are described inside the function:
https://github.com/atlassian-api/atlassian-python-api/blob/a863ad779fb7c8199d6ba037fd8189e9710ca7e7/atlassian/bitbucket/cloud/repositories/pipelines.py#L66-L69

The paging is done internal and hidden from the user:
https://github.com/atlassian-api/atlassian-python-api/blob/a863ad779fb7c8199d6ba037fd8189e9710ca7e7/atlassian/bitbucket/cloud/repositories/pipelines.py#L84-L85

I can make a suggestion for doing the workaround inside the function without user interaction.

Spacetown:
What do you think about the suggestion? Maybe the arguments `page` and `pagelen` can be removed again.

Amnesycks:
Hello @Spacetown, My bad i absolutely not think to docstrings :disappointed:. Your suggestion seems good, I don't tested it, but the logic seems better, so in my opinion we should keep yours.

Spacetown:
@Amnesycks No problem. Can you check this on your site because you had the problem?

PR #756, opened 2021-04-15
Bamboo: add environment and deployment permission

w33dw0r7d:
New methods added:

- Get deployment users
- Revoke user from deployment
- Grant user to deployment
- Get deployment groups
- Revoke group from deployment
- Grant group to deployment
- Get environment user
- Revoke user from environment
- Grant user to environment
- Get environment groups
- Revoke group from environment
-  Grant group to environment


gonchik:
Thanks for that

PR #761, opened 2021-04-22
[Bitbucket Cloud] Add avatar property to user object

RobinDeNeef:
Added an avatar property to the Bitbucket Cloud user object.


lastthyme (approved):

gonchik:
@RobinDeNeef  could you to run via black analyzer, please?

Spacetown:
This returns only the url of  the avatar. Would it be better to return the data and the content type?

PR #762, opened 2021-04-23
[Bitbucket] Pipeline paging workaround

Spacetown:
Do internal paging as suggested in #755.

Closes 740


Spacetown:
@Amnesycks Can you check this?

Amnesycks:
Hello, sorry for the other request to review I was unavailable. For this one, I done tests with my project and it seems that the pages without any pipeline still return "value" attributes but empty, so maybe a test should be added in line 67 (atlassian-python-api/atlassian/bitbucket/cloud/base.py) to prevent infinite loop. What do you think?

gonchik:
Thank you!

PR #765, opened 2021-05-04
Fix group_members() not working

altflg:
The call to get members of a certain group in bitbucket does not work because the string format method is being passed a method instead of the result of the method. This is a fix for that issue.


gonchik:
@altflg Thank you!

PR #768, opened 2021-05-06
[Bitbucket Server] Add support for getting audit log

altflg:
This was tested under Bitbucket v6.10.7 Data Center. 

The REST API documentation indicates that the Cloud version does not have an endpoint for the audit log, hence the methods get_project_audit_log() and get_repo_audit_log() will throw an Exception.


gonchik:
Thanks

Issue #597, opened 2020-10-09
Trailing slash inconsistent with Xray documentation


ghost:
There is one API call with a trailing slash which exists in
the current implementation of atlassian-python-api.

According to [the Xray documentation](https://docs.getxray.app/display/XRAY/Tests+-+REST#TestsREST-ExportingTests),
> "rest/raven/1.0/api/test/"

is incorrect and should instead be
> "rest/raven/1.0/api/test"

*I discovered this as it breaks this API call only (_rest/raven/1.0/api/test_) for my company's integration proxy, and I had to overload it as a workaround:*
```
  def get_tests(self, test_keys):
      url = 'rest/raven/1.0/api/test?keys={0}'.format(';'.join(test_keys))
      return self.get(url)
```

gonchik:
it has been fixed

PR #771, opened 2021-05-07
Removed trailing slash from api call for get_tests

micahkendall:
Per issue #597, this api call should not have a trailing slash according to Xray documentation.


gonchik:
Thank you!

PR #772, opened 2021-05-08
[Bitbucket] Add create method for repository.

### LINKED ISSUE #760: Can't create repositories in Bitbucket Cloud?
There's a method available to [create repos in Bitbucket Server](https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/bitbucket/server/projects/repos/__init__.py#L14) but I can't find an equivalent method for Bitbucket Cloud.

It seems [this update method](https://github.com/atlassian-api/atlassian-python-api/blob/a863ad779fb7c8199d6ba037fd8189e9710ca7e7/atlassian/bitbucket/cloud/repositories/__init__.py#L163) is the intended equivalent, but I can't figure out how to call this.  I thought this would be the way to do it, but the below gives an error because the instance of 'repositories' has no 'update' member.

```
from atlassian.bitbucket import Cloud

bitbucket_cloud = Cloud(
    url = 'https://api.bitbucket.org/',
    username="fake_username",
    password="password123",
    cloud=True
)

bitbucket_cloud.repositories.update(
    repo_slug="test-repo",
    workspace="AG",
    description="A test repository.",
    is_private=True
)
```

The thing that really throws me for a loop is that there's [a "delete" method](https://github.com/atlassian-api/atlassian-python-api/blob/a863ad779fb7c8199d6ba037fd8189e9710ca7e7/atlassian/bitbucket/cloud/repositories/__init__.py#L175) that's there and can be used from the same class, but not update!

Spacetown:
Implement create method mentioned in #760.

Closes #760.

### DIFF
--- a/atlassian/bitbucket/cloud/repositories/__init__.py
+++ b/atlassian/bitbucket/cloud/repositories/__init__.py
@@ -43,6 +43,8 @@ class Repositories(RepositoriesBase):
                              See https://developer.atlassian.com/bitbucket/api/2/reference/meta/filtering for details.
 
         :return: A generator for the repository objects
+
+        API docs: https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories#get
         """
         if q is not None and role is None:
             raise ValueError("Argument [q] requires argument [role].")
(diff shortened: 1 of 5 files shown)


auphofBSF:
What timing, I was just looking through the codebase for a  create repository method and wondering why there was none. Thank you for your efforts.

In my haste to use your solution to solve an immediate issue, and as your implementation is so concise , I have just copied your create method into a new class in my software that extends `WorkspaceRepositories` and then able to create a repositories in my bitbucket workspace, This justs building on my already installed pip installed atlassian-api as to not introduce dependency changes

I have however found  that I get 400 errors when I use the example from `tests/test_bitbucket_cloud.py` being `repositories.create("TestRepositoryCreated")`   When I try and use a repo name `testrepositorycreated` it works and associates correclty to the last used project.  I believe from https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D#post what you call name should be **{repo_slug}** and that must be lowercase without whitespaces and some other rules I cant lay my hands on immediately. ie `Test Repository` will generate a slug of `test-repository`

I also note that the data passed to Bitbucket does not set attributes like `is_private` or `fork_policy` . I  would prefer the `is_private` defaults to True, at present this `create` method, sets it to what the API defaults to which appears to be False. Unfortunately the bitbucket API documentation does not confirm the defaults.    Could this create method accept KWARGS that could then allow the user to set any of the permitted attributes.

Spacetown:
@auphofBSF I've renamed `name` to `repo_slug`. For setting additional properties you can use the returned object. I prefer not to change the server defaults.

auphofBSF:
> @auphofBSF I've renamed `name` to `repo_slug`. For setting additional properties you can use the returned object. I prefer not to change the server defaults.

As I get more familiar with this Atlassian API I agree that using the update method  is the consistent approach with this API wrapper. 👍 Thank you

PR #786, opened 2021-05-21
[BITBUCKET] Add support for Code Insights annotations

KevinDeJong-TomTom:
This commit will allow users to extend Code Insight reports with additional code annotations.


gonchik:
Thank you for that feature!

PR #791, opened 2021-06-04
Bitbucket: add support for downloading repository archives

sworisbreathing:
Functionality is implemented in both the low-level client as well as the resource-oriented stuff.
I've not put an equivalent in the BitBucket Cloud code because I can't find an equivalent of the /archive endpoint in the v2.0 REST API


gonchik:
Thank you!

dgarozzo:
Using your example, I get an error:

AttributeError: 'Bitbucket' object has no attribute 'download_repo_archive'

What am I doing wrong?

dgarozzo:
Looks like the pip install version doesn't have the latest code for some reason. If I download the archive from here and manually install, the download method exists.

dgarozzo:
The sample code is incorrect; the method expects dest_ft to be passed in:

with open("archive.tgz", mode="wb") as dest_fd:
	bitbucket.download_repo_archive(
		project_key="DEMO",
		repository_slug="example-repository",
		at="master",
		format="tar.gz",
		dest_fd=dest_fd
	)

gonchik:
@dgarozzo Thanks for detection, feel free to send a PR

Issue #801, opened 2021-06-29
[Bitbucket Cloud] pullrequests pagination invalid URL (using each)


rivoric:
Using pullrequests.each() only works if the number of open pull requests fits in a single page (<=10). The first call (without a page number) works but the second call fails. Debug shows the following 2 calls

> https://api.bitbucket.org:443 "GET /2.0/repositories/summitmedia/automotive-retail-frontend/pullrequests/ HTTP/1.1" 200 None
> https://api.bitbucket.org:443 "GET /2.0/repositories/summitmedia/automotive-retail-frontend/pullrequests/?page=2/ HTTP/1.1" 400 55

The problem is the trailing slash is added to the end of the URL giving the invalid page number, ie 2/

The value of next being returned from the first call is correct. The each method sets trailing=True and this causes the trailing slash to be added on subsequent calls. I think either not setting this to true or having _get_paged set trailing=False if there is next value (similar to how it sets absolute=True) would fix it.

Spacetown:
@gonchik This can be closed.

gonchik:
Thank you @Spacetown  :)

PR #803, opened 2021-07-01
[Bitbucket] Clear trailing slash in paged call of next url.

Spacetown:
Fix problem mentioned in #801.


gonchik:
Thanks as usual!

PR #804, opened 2021-07-05
[JIRA] Optimize error msg

jnhyperion:
In some cases, the http api response's "errorMessages" is empty, like this:

`{'errorMessages': [], 'errors': {'fixVersions': 'data was not an array'}}`

appending the errors to the error message is better.


gonchik:
@jnhyperion Could you run via back linter please?

Issue #604, opened 2020-10-11
[Bitbucket] Looks like the whole project is out of date for bitbucket


Mradr:
Looking over the code - and base off a fast google search - the whole project seems a bit out of date with api urls not in the correct location along with projects instead of repos or workspaces. Something is quite off here.

gonchik:
@Mradr  Do you mean about bitbucket cloud? 
Bitbucket server is up to date, I would say I use on daily basis the bitbucket server api.

Spacetown:
I also saw that the cloud is outdated. See also #540.

gonchik:
Refrshed there
https://github.com/atlassian-api/atlassian-python-api/commit/05e483ae82f01ec68d4a3d68b9d0c68d82304037

PR #810, opened 2021-07-12
Fix branches/tags for cloud usage

nborisenko:


atlassian/bitbucket/__init__.py
```diff
@@ -252,7 +252,10 @@ def upload_plugin(self, plugin_path):
     ################################################################################################
 
     def _url_projects(self, api_root=None, api_version=None):
+        if self.cloud:
+            return self.resource_url("repositories", api_root, api_version)
```
Spacetown, line 256: This doesn*t make sense for me. Why is the URL for `projects` redirected `repositories`???

nborisenko: Because this how api looks like in cloud. /projects route not supported in cloud.
https://developer.atlassian.com/bitbucket/api/2/reference/search?q=projects
https://developer.atlassian.com/bitbucket/api/2/reference/search?q=refs

Spacetown: Thats because here it is called `workspace`. There is an open issue #604. If this API is needed I suggest to add it into the `Cloud` part of BitBucket.

nborisenko: I agreed with you, from my point of view it should be 2 separate clients. But I currently didn`t have enough time for move it to cloud class or build new client. This is temporary solution, which allows persons like me use python client with bitbucket.org. Because it`s fully unusable in case when you want to do something with commits/tags/branches in cloud.

gonchik:
Closing that PR as we have dedicated client for server

PR #816, opened 2021-07-13
[JIRA] Documentation and example addition for getting all projects

isaac-philip:
Straightforward documentation and example addition, that's all.


gonchik:
thanks

PR #795, opened 2021-06-09
Jira: Update assign_issue to match the current API

Sebelino:
Fixes https://github.com/atlassian-api/atlassian-python-api/issues/515 and addresses the comments left in https://github.com/atlassian-api/atlassian-python-api/pull/728.

PR #821, opened 2021-07-17
Assign Issue Fix

### LINKED ISSUE #775: Need example on how to assign issue
This project could sure use more examples of how to do even basic things.

For example, I'd like to assign an issue being created to the current user.  I have no idea how to do that...
No example exists to show how to assign to another users either...

Maybe someone could point to working examples elsewhere?  I'd be happy to do a couple of PRs for examples once I figure out how to do some of this stuff.

isaac-philip:
- Changes to existing Assign Issue to user code

- Example added

- taken from https://docs.atlassian.com/software/jira/docs/api/REST/8.5.0/#api/2/issue-assign

Resolves #775

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -973,7 +973,10 @@ class Jira(AtlassianRestAPI):
         """
         base_url = self.resource_url("issue")
         url = "{base_url}/{issue}/assignee".format(base_url=base_url, issue=issue)
-        data = {"accountId": account_id}
+        data = {"name": account_id}
+        log.warning(
+            "Assigning Issue {issue_key} to User Id {account_id}".format(issue_key=issue, account_id=account_id)
+        )
         return self.put(url, data=data)
 
     def create_issue(self, fields, update_history=False, update=None):
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -974,9 +974,6 @@ class Jira(AtlassianRestAPI):
         base_url = self.resource_url("issue")
         url = "{base_url}/{issue}/assignee".format(base_url=base_url, issue=issue)
         data = {"name": account_id}
-        log.warning(
-            "Assigning Issue {issue_key} to User Id {account_id}".format(issue_key=issue, account_id=account_id)
-        )
         return self.put(url, data=data)
 
     def create_issue(self, fields, update_history=False, update=None):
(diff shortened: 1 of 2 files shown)


atlassian/jira.py
```diff
@@ -973,7 +973,10 @@ def assign_issue(self, issue, account_id=None):
         """
         base_url = self.resource_url("issue")
         url = "{base_url}/{issue}/assignee".format(base_url=base_url, issue=issue)
-        data = {"accountId": account_id}
+        data = {"name": account_id}
+        log.warning(
```
Spacetown, line 977: This warning should be removed.

Spacetown (changes requested):

Spacetown:
It seems that Atlassion is toggling the parameters. #728 was fixed by #795 and now the change is reverted.

atlassian/jira.py
isaac-philip: Sure @Spacetown , but before could you help with the reason please ?

Just following the similar convention used in the other functions within same `jira` module like,

1. https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/jira.py#L958

2. https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/jira.py#L1031

and many more.

warning is lower logger level than info, so shouldn't be noise as per my knowledge.

If we have a conflict, would like @gonchik to suggest a way forward.

Spacetown: Please consider https://docs.python.org/3/howto/logging.html:
- warning: `Issue a warning regarding a particular runtime event`
- info: `Report events that occur during normal operation of a program (e.g. for status monitoring or fault investigation)`

In normal operation no messages are good messages.

isaac-philip: Ok thanks for the reply,  I agree to what your referring to but your missing the point I am making. 😄 

I didn't get your reply to my query for existing warning logs placed. Could you help out for that please ?

Is `issue_add_watcher` a normal operation? - https://github.com/atlassian-api/atlassian-python-api/blame/master/atlassian/jira.py#L958
You could explain this line especially since you added it.

I will happily clear this but then does it make sense to have the existing warning logs also removed and from which functions I am unclear on? seems a select few? please help on those too with an Issue then.

@gonchik could you help out here please if this conversation is taking too long 😆

Spacetown: @isaac-philip warning is more noise than info. Warnings and errors are visible per default, Infos are only activated in verbose mode..
I'm with you that the other places should also be to changed.
If the function is called the user is warned that the desired action is done but this is the intention.

Spacetown (approved):

gonchik:
Thanks @isaac-philip  @Spacetown  for collaborative work :)

PR #830, opened 2021-08-03
[BitBucket] Replace fstring with format function.

### LINKED ISSUE #828: version 3.12 introduced a small python < 3 issue by using an fstring
in the newest version line 2638 in __init__.py for bitbucket
url = f"{self._url_repo(project_key, repository_slug)}/archive"
uses an fstring which are not  supported in python <3

I used 
url = self._url_repo(project_key, repository_slug)+"/archive"
instead

Spacetown:
@gonchik  Can the usage of f-strings be checked with flake8? I found nothing.

Closes #828


Spacetown:
I also find only checks if a fostering has no code references. Thanks.

PR #836, opened 2021-08-09
Bitbucket: add webhook test

### LINKED ISSUE #808: unit test for webhook
Hi @gonchik 

I noticed you have #806 merged. Thank you.

As we just invoke the original rest api  #806, the unit tests would be too simple.

Do we need to add this kind of code style of unit test for this feature?
If yes, I'd like to summit another PR.
If no, could you please show some better code style of unit test?

Here is the snippet. It seems stupid, but it does work.
```
from atlassian import Bitbucket
from unittest.mock import patch
from unittest import TestCase

class TestWebhook(TestCase):

    def setUp(self):
        self.fake_data = {
            "project_key": "fake_project",
            "repository_slug": "fake_repo",
            "name": "fake_name",
            "events": ["repo:refs_changed", "pr:merged", "pr:opened"],
            "webhook_url": "https://example.com",
            "active": True,
            "secret": "fake_secret",
        }

    def test_create_webhook(self):
        return_value = {
            "id": 10,
            "name": "fake_name",
            "createdDate": 1513106011000,
            "updatedDate": 1513106011000,
            "events": [
                "repo:refs_changed",
                "pr:merged",
                "pr:opened"
            ],
            "configuration": {
                "secret": "fake_secret"
            },
            "url": "https://example.com",
            "active": True
        }
        with patch.object(Bitbucket, "post", return_value=return_value):
            bitbucket = Bitbucket(url="https://bitbucket.example.com", username="username", password="password")
            rst = bitbucket.create_webhook(self.fake_data["project_key"], self.fake_data["repository_slug"],
                                           self.fake_data["name"], self.fake_data["events"],
                                           self.fake_data["webhook_url"], self.fake_data["active"])
            self.assertEqual(rst["name"], self.fake_data["name"])
            self.assertEqual(rst["events"], self.fake_data["events"])
            self.assertEqual(rst["configuration"]["secret"], self.fake_data["secret"])
            self.assertEqual(rst["url"], self.fake_data["webhook_url"])
            self.assertEqual(rst["active"], self.fake_data["active"])

```

zhan9san:
Fix #808

Add webhook test

Could you kindly review and approve this PR?


Spacetown (commented):
Maybe it would be better to split the tests for the OO interface and the old interface into separate files?

tests/test_bitbucket_server_webhook.py
```diff
@@ -0,0 +1,95 @@
+from unittest import TestCase
+
+from atlassian.bitbucket import Bitbucket
+
+from .mockup import mockup_server
```
zhan9san, line 5: Hi @Spacetown 

The mockup_server ImportError is caught in `test_bitbucket_cloud.py`, `test_bitbucket_server.py` and `test_servicedesk.py`.

IMHO, once this error is passed, the object would be None, and the following test must fail.
So I remove the try...except statement. Is it ok?

Is try...except designed for another reason?

zhan9san:
Referring to `test_confluence_attach.py`, the test file name is restricted in the `_webhook` suffix.

Is it better to create an individual test file for subsequent test of other resource, or you prefer a common name this time and add new test in this common test file?

tests/test_bitbucket_server_webhook.py
Spacetown: No, the test will not fail because the is skipped if python < 3.4 where the mockup isn't possible:
```python
@pytest.mark.skipif(sys.version_info < (3, 4), reason="requires python3.4")
```

zhan9san: Oh, yes. It will succeed if python < 3.4.
Is this the only reason to introduce try...except?

Spacetown: Yes. @gonchik the test don't use this python version anymore. Can this be removed or should it be kept for users developing on python 2.7?

Spacetown:
What about renaming `test_bitbucket_server.py` to `test_bitbucket_server_oo.py` and `test_bitbucket_cloud.py` to `test_bitbucket_cloud_oo.py` and add your test and other tests for the non OO implementation to `test_bitbucket_server.py`?

zhan9san:
> What about renaming `test_bitbucket_server.py` to `test_bitbucket_server_oo.py` and `test_bitbucket_cloud.py` to `test_bitbucket_cloud_oo.py` and add your test and other tests for the non OO implementation to `test_bitbucket_server.py`?

Hi @Spacetown 

Finally I got time to work on it. All files above have been renamed.

BTW, please allow me ask some stupid questions, as I don't have much experience
What does OO mean here? Is it Object Oriented?
How do we differentiate these test cases?

Spacetown:
Questions are never stupid. Yes it is object oriented.

Issue #839, opened 2021-08-12
[Bitbucket Cloud] Query string not working on page > 1


mahdidul:
HI,

I'm trying to use query string to get list of merged pullrequests in a repository : 

```
bitbucket = Cloud(username=u, password=password)

    w = bitbucket.workspaces.get(workspace)
    p = w.projects.get(project)
    for r in p.repositories.each():
        pr_ids = []
        for pr in r.pullrequests.each(q='state="MERGED"'):
            pr_ids.append(pr.id)
        print(r.name + " : " + str(pr_ids))
```

But the url requests seems to be messed up with that query string and returns 400 - Bad Request as seen below

```
Traceback (most recent call last):
  File "C:\Users\xx\PycharmProjects\pythonProject\bitbucket_report.py", line 22, in <module>
    for pr in r.pullrequests.each(q='state="MERGED"'):
  File "C:\Users\xx\anaconda3\envs\pythonProject\lib\site-packages\atlassian\bitbucket\cloud\repositories\pullRequests.py", line 74, in each
    for pr in self._get_paged(None, trailing=True, params=params):
  File "C:\Users\xx\anaconda3\envs\pythonProject\lib\site-packages\atlassian\bitbucket\cloud\base.py", line 59, in _get_paged
    response = super(BitbucketCloudBase, self).get(
  File "C:\Users\xx\anaconda3\envs\pythonProject\lib\site-packages\atlassian\rest_client.py", line 264, in get
    response = self.request(
  File "C:\Users\xx\anaconda3\envs\pythonProject\lib\site-packages\atlassian\rest_client.py", line 236, in request
    self.raise_for_status(response)
  File "C:\Users\xx\anaconda3\envs\pythonProject\lib\site-packages\atlassian\rest_client.py", line 386, in raise_for_status
    response.raise_for_status()
  File "C:\Users\xx\anaconda3\envs\pythonProject\lib\site-packages\requests\models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.bitbucket.org/2.0/repositories/xx/xx-xx/pullrequests/?q=state%3D%22MERGED%22&page=2/&q=state%3D%22MERGED%22

Process finished with exit code 1
```

mrtalkless:
Seems the issue lies in the next page (page > 1) since the first page was success as seen in the Output below
```
[...]

PR #840, opened 2021-08-12
 [BitBucket] Add parameter Trailing on each function.

mahdidul:
The trailing url caused issue #839 if there's query string on `each` function when iterating on `next` page.
This PR will provide an option to override `Trailing` parameter with the default value `Trailing=True` to keep backward compatibility.


Spacetown (changes requested):
That's not the cause of your problem. Please take a look at #841

PR #843, opened 2021-08-17
[Bitbucket] Adding ability to fork a repo to a different project

zacaryc:
Adding ability to fork a repo to a different project that differs from the original source project, which is what the existing
functionality is limited to.

Within a fair few of my workflows and projects I work with, we need to fork the original project to a personal user repo make changes and create pull requests back. Unfortunately in trying to automate these I'm limited in the ability of using this library to be able to bootstrap the workflow.

This is doing the bare minimum fork without adding in any extra properties like fork syncing etc to add the smallest change first that matches the existing mono-project fork.

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -855,6 +855,23 @@ class Bitbucket(BitbucketBase):
             body["project"] = {"key": project_key}
         return self.post(url, data=body)
 
+    def fork_repository_new_project(self, project_key, repository_slug,
+                                    new_project, new_repository_slug):
+        """
+        Forks a repository to a separate project.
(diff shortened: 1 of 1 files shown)


zacaryc:
Noticing that every PR that's raised recently seems to be failing the Test run on the same step.
Is it stable/failing on master already and these are false negatives? Or is there something missing here to be added to test code, for my PR?

gonchik:
Thanks! :) 
Please, provide for the next time a docs.

zacaryc:
@gonchik thank you. 
No worries I'll write up some docs now actually, I noticed the regular fork was missing them as well.

zacaryc:
https://github.com/atlassian-api/atlassian-python-api/pull/852 created for docs

PR #846, opened 2021-08-21
Respect `advanced_mode` in calls of jql.

### LINKED ISSUE #835: Get issues of a non exist projetc
Hi,
I'm trying to get all issues of a project, the problem is that if it doesn't exist, I can't send the response that says "project doesn't exist or you don't have permissions", I only can send "400 Bad Request".
I see this message in log file, but I can't catch it to be used as response.

Thanks

Spacetown:
Closes #835


gonchik:
Thanks! :)

Issue #847, opened 2021-08-26
Unable to trigger a specific pipeline without providing commit key 


arvind-kalyan:
Hello, I'm trying out the bitbucket pipelines python api, but running into an issue where it mandatorily asks for `commit` to be non None while calling pipelines.trigger() API.

What I'm trying to do is to be able to trigger a custom pipeline for a branch : `r.pipelines.trigger(branch="master", pattern="deployment-to-test-infra", variables=[{'key': 'SERVER_ENV', 'value': 'test-01'}) `, it's giving the error : `ValueError: Missing argument [commit].`. 

I'm however able to trigger the HTTP API directly without commits and it works fine.  
```
        if pattern is not None:
            if commit is None:
                raise ValueError("Missing argument [commit].")
            data["target"]["selector"] = {
                "type": "custom",
                "pattern": pattern,
            }
```
I'm not sure why this inner IF check was added. I can help with the PR if you can let me know the changes required.

Spacetown:
@gonchik This can be closed.

PR #848, opened 2021-08-26
[bitbucket pipelines] allow commit arg to be optional when only pattern is passed in pipeline trigger api

arvind-kalyan:
Fixes Issue : https://github.com/atlassian-api/atlassian-python-api/issues/847


gonchik:
Thanks!

PR #853, opened 2021-09-05
[Bitbucket] Removing redundant double check for simpler code to read (minor/readability)

zacaryc:
Two concurrent if statements that check the same thing, instead of double checking the code, just nesting both statements within the same identical conditional


gonchik:
Got it! Thanks!

PR #859, opened 2021-09-19
Added parameters to projects get request

EgorDm:
Adds missing parameter argument for projects get request


gonchik:
@EgorDm Thanks for catching that :)

PR #844, opened 2021-08-18
[Bitbucket] Fix pagination for cloud requests

sergey-p3:
Exception raised when requesting pull requests which exceed single page: requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.bitbucket.org/2.0/repositories/XXXX/pullrequests/?q=state+%3D+%22OPEN%22+OR+state+%3D+%22MERGED%22+or+state%3D%22DECLINED%22&page=2/&q=state+%3D+%22OPEN%22+OR+state+%3D+%22MERGED%22+or+state%3D%22DECLINED%22

Next argument returned by the response contains full url including original params and next page. The issue is that the params are passed as duplicates to the following request along with the trailing / after provided url.

PR #861, opened 2021-09-29
Fix missing data when using pullRequests.each()

sti0:
This fixes a bug that ends in missing data when using `pullRequests.each()`.

The pullrequests endpoint won't deliver the 
whole PullRequest object (e.g. participants, reviewers missing).
So we need to load all the data for the PR object.

The [docs](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pullrequests) say:
> The list of users that were added as reviewers on this pull request when it was created. **For performance reasons, the API only includes this list on a pull request's self URL.**

After merging this PR the PullRequest API is still broken and needs to be fixed by #844


gonchik:
Thanks!

PR #863, opened 2021-10-05
[Confluence] Fixed template functions urls using wiki/rest/.

olaor:
All other functions in confluence.py use rest/api/.

confluence.get_content_templates() works a lot better now.


Mooreel (approved):
As the init always adds "/wiki" to the url, it is correct that the calls shouldn't contain "wiki" in the URL. 

This pr fixes this.

gonchik:
Thanks

PR #889, opened 2021-11-21
Fix delete watchers

jensvog:
Unlike in add watcher, the user parameter has to be handed over as query parameter, according to the documentation: https://docs.atlassian.com/software/jira/docs/api/REST/8.13.14/#issue-removeWatcher. This patch fixes this issue.


gonchik:
Thanks

PR #903, opened 2021-12-23
Methods for Jira filter share permissions

jensvog:
Adds these methods:
* Get Jira filter share permissions
* Add Jira filter share permission
* Delete Jira filter share permission


gonchik:
Thank you!

PR #904, opened 2021-12-24
Add exponential backoff and retry functionality to rest_client

jp-harvey:
While working with the Atlassian Cloud API and attempting to optimize the time it takes to do many operations, it's common to hit the [Atlassian API request limits](https://developer.atlassian.com/cloud/jira/platform/rate-limiting/), resulting in annoying negative engineering.

This PR adds a convenience option to the underlying REST client, turned off by default, to retry using exponential backoff.


jp-harvey:
@gonchik I've not tested this quite enough for it not to be a `[WIP]`, will update after I've done so.

gonchik:
@jp-harvey  checked a few options, that's very great step for the rest client module  :) 
Thank you for your initiative, I hope soon it will be done.

atlassian/rest_client.py
```diff
@@ -52,7 +53,45 @@ def __init__(
         cloud=False,
         proxies=None,
         token=None,
+        backoff_and_retry=False,
+        retry_error_matches=[(429, "Too Many Requests")],
```
flichtenheld, line 57: Tried this with Bitbucket Cloud today. Required the following change to make it work:
```suggestion
        retry_error_matches[(429, "Too Many Requests"),
                                         (429, "Unknown Status Code")],
```

flichtenheld:
Another change I made was to add these new options to `BitbucketBase._new_session_args`. At least my assumption was that if you set backoff_and_retry=True when creating the `Cloud` object you would expect all child objects to inherit the setting.

jp-harvey:
Great suggestion @flichtenheld , I've been working with Jira and not with Confluence so thanks for adding that in, glad it helped you out :-)

jp-harvey:
> Another change I made was to add these new options to `BitbucketBase._new_session_args`. At least my assumption was that if you set backoff_and_retry=True when creating the `Cloud` object you would expect all child objects to inherit the setting.

@flichtenheld do you want to add those as a review to this PR as well?

flichtenheld:
> > Another change I made was to add these new options to `BitbucketBase._new_session_args`. At least my assumption was that if you set backoff_and_retry=True when creating the `Cloud` object you would expect all child objects to inherit the setting.
> 
> @flichtenheld do you want to add those as a review to this PR as well?

Hmm, not sure what would be the best way to do that given that they are in a completely different file that is not touched by the PR. But you can see them here: https://github.com/flichtenheld/atlassian-python-api/commit/f94cef60fd313bfb6721e921c0f4609b71114046

atlassian/rest_client.py
```diff
@@ -221,17 +265,36 @@ def request(
...
+            responseloop = False
+            if self.backoff_and_retry:
+                if retries > self.max_backoff_retries:
+                    log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                    responseloop = False
+                for em in self.retry_error_matches:
+                    if response.status_code == em[0] and response.reason == em[1]:
+                        log.warning('Backing off due to error "{0}: {1}" for {2}s'.format(em[0], em[1], backoff))
+                        time.sleep(backoff)
+                        backoff = backoff * 2 if backoff * 2 < self.max_backoff_seconds else self.max_backoff_seconds
```
kkosciusz, line 293: backoff = min(2 * backoff, self.max_backoff_seconds)

atlassian/rest_client.py
```diff
@@ -221,17 +265,36 @@ def request(
...
+                url=url,
+                headers=headers,
+                data=data,
+                json=json,
+                timeout=self.timeout,
+                verify=self.verify_ssl,
+                files=files,
+                proxies=self.proxies,
+            )
+            responseloop = False
```
Spacetown, line 284: Setting always to `False` and in Line 295 to `True` is confusing. Remove line 284 and 295.

atlassian/rest_client.py
```diff
@@ -221,17 +265,36 @@ def request(
...
+            if self.backoff_and_retry:
+                if retries > self.max_backoff_retries:
+                    log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                    responseloop = False
+                for em in self.retry_error_matches:
+                    if response.status_code == em[0] and response.reason == em[1]:
+                        log.warning('Backing off due to error "{0}: {1}" for {2}s'.format(em[0], em[1], backoff))
+                        time.sleep(backoff)
```
Spacetown, line 292: If `Retry-After` is send in the Header this should be used, see https://developer.atlassian.com/cloud/jira/platform/rate-limiting/#retry-after-headers.

Spacetown, line 292: If `Retry-After` is send in the Header this should be used, see https://developer.atlassian.com/cloud/jira/platform/rate-limiting/#retry-after-headers.

atlassian/rest_client.py
```diff
@@ -221,17 +265,36 @@ def request(
...
+                timeout=self.timeout,
+                verify=self.verify_ssl,
+                files=files,
+                proxies=self.proxies,
+            )
+            responseloop = False
+            if self.backoff_and_retry:
+                if retries > self.max_backoff_retries:
+                    log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                    responseloop = False
```
Spacetown, line 288: This block should be moved inside the `for` loop. Else the warning is issued and if the response is a failure there will be an additional sleep and then the loop is terminated.

Spacetown, line 288: This block should be moved inside the `for` loop. Else the warning is issued and if the response is a failure there will be an additional sleep and then the loop is terminated.

Spacetown (changes requested):

atlassian/rest_client.py
jp-harvey: @Spacetown this is a great suggestion. The way it was implemented is based on the same article further down: https://developer.atlassian.com/cloud/jira/platform/rate-limiting/#request-backoff

> Request backoff

> Apps should treat 429 responses as a signal to alleviate pressure on an endpoint. Apps should retry the request only after a delay. Best practice to double the delay after each successive 429 response from a given endpoint. Backoff delays only need to exponentially increase to a maximum value at which point the retries can continue with the fixed maximum delay. You should also apply [jitter](https://en.wikipedia.org/wiki/Jitter)
to the delays to avoid the [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem)

100% agree a more complete implementation would include the `Retry-After`. The exponential backoff and retry covers all scenarios whether the header is returned or not so is the minimum viable functionality. At present I don't have capacity to add the `Retry-After` logic and for this version I'd suggest it's unneeded functionality that would delay this PR being merged. This suggestion would be a great GH issue though for future improvement.

atlassian/rest_client.py
jp-harvey: Good one @kkosciusz, I like your version better :slightly_smiling_face:

atlassian/rest_client.py
jp-harvey: @Spacetown I think that would cause an infinite loop if `self.backoffandretry` is set to `False`, which is the default.

If it's confusing is there another way we could implement it?

Now that you mention it I'm also just looking at line 288 and I think that could probably be removed although I think I put that in there to future proof any other logic changes relating to `responseloop` that might change the default - ie. making the logic explicit - so would probably vote to leave that in.

atlassian/rest_client.py
jp-harvey: (resolving duplicate)

atlassian/rest_client.py
jp-harvey: (duplicate comment)

atlassian/rest_client.py
jp-harvey: @Spacetown forgot `break` does not apply to `if` statements :slightly_smiling_face:, moved as per your suggestion and added a `break`

jp-harvey:
@gonchik there has been no additional feedback by reviewers for some time following changes / responses and I think this one is probably good to go?

Spacetown (commented):
Oops, forgot to submit the review.

atlassian/rest_client.py
```diff
@@ -221,17 +265,37 @@ def request(
...
+                verify=self.verify_ssl,
+                files=files,
+                proxies=self.proxies,
+            )
+            responseloop = False
+            if self.backoff_and_retry:
+                for em in self.retry_error_matches:
+                    if retries > self.max_backoff_retries:
+                        log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                        responseloop = False
```
Spacetown, line 289: This line isn't needed because you use break to exit the loop.

atlassian/rest_client.py
```diff
@@ -221,17 +265,37 @@ def request(
...
+                json=json,
+                timeout=self.timeout,
+                verify=self.verify_ssl,
+                files=files,
+                proxies=self.proxies,
+            )
+            responseloop = False
+            if self.backoff_and_retry:
+                for em in self.retry_error_matches:
+                    if retries > self.max_backoff_retries:
```
Spacetown, line 287: This block should be moved in front of the loop.

atlassian/rest_client.py
jp-harvey: Jitter done

gonchik:
@jp-harvey @flichtenheld  can we update PR and merge it ?

jp-harvey:
> @jp-harvey @flichtenheld can we update PR and merge it ?

@gonchik I've made some of the changes, the others I am not comfortable making without doing testing and I don't have a good way to test at the moment. Functionally it works as it is and has been tested many times in real-life scenarios, it's just the code structure that is the concern now.

Spacetown:
> > @jp-harvey @flichtenheld can we update PR and merge it ?
> 
> @gonchik I've made some of the changes, the others I am not comfortable making without doing testing and I don't have a good way to test at the moment. Functionally it works as it is and has been tested many times in real-life scenarios, it's just the code structure that is the concern now.
> 
> EDIT: looks like there's a linting issue as well which should be easy enough to sort out.

It works but it's wired and not easy to understand. If max back off is reached but the response was success a warning is logged.
```python
        backoff = 1
        retries = 0
        responseloop = True
        while responseloop:
            response = self._session.request(
                method=method,
                url=url,
                headers=headers,
                data=data,
                json=json,
                timeout=self.timeout,
                verify=self.verify_ssl,
                files=files,
                proxies=self.proxies,
            )
            if self.backoff_and_retry:
                for em in self.retry_error_matches:
                    if response.status_code == em[0] and response.reason == em[1]:
                        if retries > self.max_backoff_retries:
                            log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
                            responseloop = False
                        else:
                            log.warning('Backing off due to error "{0}: {1}" for {2}s'.format(em[0], em[1], backoff))
                            time.sleep(backoff)
                            backoff = backoff * 2 if backoff * 2 < self.max_backoff_seconds else self.max_backoff_seconds
                            retries += 1
                        break. # for loop
            else:
                responseloop = False

```

Spacetown:
The for loop and the if can also be combined using `any()` together with comparing `set` objects.

atlassian/rest_client.py
```diff
@@ -60,7 +60,46 @@ def __init__(
         cloud=False,
         proxies=None,
         token=None,
+        backoff_and_retry=False,
+        retry_error_matches=[(429, "Too Many Requests"),
+                                         (429, "Unknown Status Code")],
```
Spacetown, line 65: Fix indention.

atlassian/rest_client.py
```diff
@@ -236,17 +279,37 @@ def request(
...
+                verify=self.verify_ssl,
+                files=files,
+                proxies=self.proxies,
+            )
+            responseloop = False
+            if self.backoff_and_retry:
+                for em in self.retry_error_matches:
+                    if retries > self.max_backoff_retries:
+                        log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                        responseloop = False
```
Spacetown, line 303: This line isn't needed since it's set to false directly after the response.

atlassian/rest_client.py
```diff
@@ -236,17 +279,37 @@ def request(
...
-            timeout=self.timeout,
-            verify=self.verify_ssl,
-            files=files,
-            proxies=self.proxies,
-        )
+
+        backoff = 1
+        retries = 0
+        responseloop = True
+        while responseloop:
```
Spacetown, line 286: Make a endless loop and use break instead of mixing both.
```suggestion
        while True:
```

atlassian/rest_client.py
```diff
@@ -236,17 +279,37 @@ def request(
...
+                url=url,
+                headers=headers,
+                data=data,
+                json=json,
+                timeout=self.timeout,
+                verify=self.verify_ssl,
+                files=files,
+                proxies=self.proxies,
+            )
+            responseloop = False
```
Spacetown, line 298:
```suggestion
```

atlassian/rest_client.py
```diff
@@ -236,17 +279,37 @@ def request(
...
+                        log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                        responseloop = False
+                        break
+                    if response.status_code == em[0] and response.reason == em[1]:
+                        log.warning('Backing off due to error "{0}: {1}" for {2}s'.format(em[0], em[1], backoff))
+                        time.sleep(backoff + (random.random() * backoff / 10))
+                        backoff = min(2 * backoff, self.max_backoff_seconds)
+                        retries += 1
+                        responseloop = True
+                        break
```
Spacetown, line 311:
```suggestion
            else:
                break
```

Spacetown (commented):
I would use only break to finish the loop. At the moment this is mixed and in line 310-311 the flag sets a new iteration but a break is used which doesn't make sense.

Spacetown:
Ok, now I see that there are two nested loops...

atlassian/rest_client.py
```diff
@@ -236,17 +279,35 @@ def request(
...
+                        log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
+                        responseloop = False
+                        break
+                    if response.status_code == em[0] and response.reason == em[1]:
+                        log.warning('Backing off due to error "{0}: {1}" for {2}s'.format(em[0], em[1], backoff))
+                        time.sleep(backoff + (random.random() * backoff / 10))
+                        backoff = min(2 * backoff, self.max_backoff_seconds)
+                        retries += 1
+            else:
+                break
```
Spacetown, line 309:
```suggestion
        response_loop = True
        while response_loop:
            response = self._session.request(
                method=method,
                url=url,
                headers=headers,
                data=data,
                json=json,
                timeout=self.timeout,
                verify=self.verify_ssl,
                files=files,
                proxies=self.proxies,
            )
            if self.backoff_and_retry:
                for em in self.retry_error_matches:
                    if retries > self.max_backoff_retries:
                        log.warning("Hit max backoff retry limit of {0}, no more retries.".format(self.max_backoff_retries))
                        responseloop = False
                        break
                    if response.status_code == em[0] and response.reason == em[1]:
                        log.warning('Backing off due to error "{0}: {1}" for {2}s'.format(em[0], em[1], backoff))
                        time.sleep(backoff + (random.random() * backoff / 10))
                        backoff = min(2 * backoff, self.max_backoff_seconds)
                        retries += 1
            else:
                response_loop = False
```

Spacetown (changes requested):
The suggested code only sets `response_loop` to False if the loop shall be exited.

gonchik:
@jp-harvey  can we finalize it ?

jp-harvey:
> @jp-harvey can we finalize it ?

Hi @gonchik I'm at a point where it's no longer 100% clear to me what the issues are or what to change. I also don't have a good way to test at the moment, so making changes and testing is going to be high-friction.

I've also discovered subsequently that the `requests` library has a[ built-in backoff and retry capability built into sessions](https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries), so that's probably a better option all round.

I think we have the following options:
1. Abandon this PR
2. Merge it (mostly?) as-is
3. Someone other than me modifies the code, tests, and once approved it can be merged
4. I modify the existing code, test, and once approved it gets merged
5. I resubmit the PR using the retry capability of `sessions`

Unfortunately I won't have the bandwidth to do it for a while and don't have an easy way to test it at the moment, so 4 and 5 won't be able to happen soon. 

Sorry about this, I know the code works as it is now despite not being perfect, and I respect that it's considered not suitable for merging. By the time the PR was reviewed we passed a window I was able to keep working on it and don't want to make blind changes now that may break it. I may be able to in the future though, but would probably do option 5 before 4 because it looks like a more elegant option. I'm also fine with options 1-3 if that's the best thing for this project. Let me know what your preference is.

atlassian/rest_client.py
mjurbanski-reef, line 65: 503 also deserve a retry IMO, since jira cloud does throw it from time to time

posting it here if someone want to update this PR or create a new one using requests.Session retry mechanism

flichtenheld: I included 503 in my new PR #1339 since that is in the default list of urllib3.util.Retry

flichtenheld:
Since #1339 was merged, I think this PR should be closed.

jp-harvey:
@flichtenheld thanks for taking this one up and your great work on https://github.com/atlassian-api/atlassian-python-api/pull/1339!

PR #906, opened 2021-12-29
Added issue_restore method

SLRover:
Added issue_restore method, which restores an issue from archive. Updated docs.


gonchik:
Thank you, mate!

PR #910, opened 2022-01-04
[JIRA] added jira oauth2 integration to examples

sureshdsk:
- added example code to implement jira cloud oauth2 integration


gonchik:
Thanks! :)

PR #912, opened 2022-01-04
add notify users flag to create issue

hansonmcoombs:
simple extension of the same method used for edit_issue to the create_issue method

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -1092,7 +1092,7 @@ class Jira(AtlassianRestAPI):
         data = {"name": account_id}
         return self.put(url, data=data)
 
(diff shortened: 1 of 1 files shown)


gonchik:
Thanks

hansonmcoombs:
I’m sorry to say this, but I made a mistake and you should roll back this
merge… that feature is not implemented in the jira rest api, but does not
raise any exceptions. Also my users said they did not receive any emails…
but actually did not check! 🤦‍♂️

There is a support ticket for this feature.
https://jira.atlassian.com/browse/JRASERVER-67363 I’ll try to progress this
instead.

Thanks for your patience.

Issue #917, opened 2022-01-06
What is the status of support in Python 3.10?


mbk4:
Hi!
As I understand from this page [https://pypi.org/project/atlassian-python-api/3.17.0/](https://pypi.org/project/atlassian-python-api/3.17.0/) Python up to maximum of 3.9 is supported. Is the Python 3.10 support planned in the near future?
Thank you for the responses.

gonchik:
Hi! Let me add it

mbk4:
> Hi! Let me add it https://docs.python.org/3/whatsnew/3.10.html

@gonchik Thank you for the reply. So this library is supported in Python 3.10 and it is just the badge missing?

gonchik:
@mbk4 I would say yes.

gonchik:
Now it works well.

PR #918, opened 2022-01-07
Add 3.10 python support badge and docker file

gonchik:
Adding support the next app:
#917


gonchik:
Something wrong with distutils, it should be warning as deprecated, not did not work

PR #925, opened 2022-01-19
Bitbucket Cloud: provide custom raise_for_status

flichtenheld:
The default one in AtlassianRestAPI does not work
for Bitbucket Cloud.

Signed-off-by: Frank Lichtenheld <frank@lichtenheld.com>

### DIFF
--- a/atlassian/bitbucket/cloud/base.py
+++ b/atlassian/bitbucket/cloud/base.py
@@ -2,6 +2,8 @@
 
 from ..base import BitbucketBase
 
+from requests import HTTPError
+
 
 class BitbucketCloudBase(BitbucketBase):
     def __init__(self, url, *args, **kwargs):
@@ -84,3 +86,27 @@ class BitbucketCloudBase(BitbucketBase):
                 trailing = False
 
         return
+
+    def raise_for_status(self, response):
+        """
+        Checks the response for errors and throws an exception if return code >= 400
+
+        Implementation for Bitbucket Cloud according to
+        https://developer.atlassian.com/cloud/bitbucket/rest/intro/#standardized-error-responses
+
+        :param response:
+        :return:
+        """
+        if 400 <= response.status_code < 600:
+            try:
+                j = response.json()
+                e = j["error"]
+                error_msg = e["message"]
+                if e.get("detail"):
+                    error_msg += "\n" + e["detail"]
+            except Exception:
+                response.raise_for_status()
+            else:
+                raise HTTPError(error_msg, response=response)
+        else:
+            response.raise_for_status()


flichtenheld:
Found this one while trying to make #904 work for Bitbucket Cloud.

salmannotkhan:
It'll be better if we just revert https://github.com/atlassian-api/atlassian-python-api/commit/50b5c71c05fa31f14dc9dcf1e3b89b88ce56df58 instead of adding error handling one by one

salmannotkhan:
> > It'll be better if we just revert [50b5c71](https://github.com/atlassian-api/atlassian-python-api/commit/50b5c71c05fa31f14dc9dcf1e3b89b88ce56df58) instead of adding error handling one by one
> 
> What was the behavior before? To me it looks like before it just used the HTTP error message which is useless in many cases. It is definitely better to show the actual error message from the JSON response. But it seems that the JSON format is different for each service...

If you look the tagged commit carefully you'll see it was showing proper error message from json data.

flichtenheld:
> > > It'll be better if we just revert [50b5c71](https://github.com/atlassian-api/atlassian-python-api/commit/50b5c71c05fa31f14dc9dcf1e3b89b88ce56df58) instead of adding error handling one by one
> > 
> > 
> > What was the behavior before? To me it looks like before it just used the HTTP error message which is useless in many cases. It is definitely better to show the actual error message from the JSON response. But it seems that the JSON format is different for each service...
> 
> If you look the tagged commit carefully you'll see it was showing proper error message from json data. Also it has different error handling for each service.

That might be true for some services, but that commit did not remove any error handling from `atlassian/bitbucket`. I just verified that. Without my patch and with the referenced commit reverted you get something like (for a random error that I hit today):

    requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.bitbucket.org/2.0/repositories/openvpntechnologies/ovpn3-build/commit/4e857d6c9df6956932ed0b25abb8454892dcebea/statuses/build

With my patch you get:

    requests.exceptions.HTTPError: Authenticated user is neither the creator (frank_lichtenheld) of this build status nor an admin of this repository.

So I think this is a definite improvement for Bitbucket Cloud specifically.

salmannotkhan:
> > > > It'll be better if we just revert [50b5c71](https://github.com/atlassian-api/atlassian-python-api/commit/50b5c71c05fa31f14dc9dcf1e3b89b88ce56df58) instead of adding error handling one by one
> > > 
> > > 
> > > What was the behavior before? To me it looks like before it just used the HTTP error message which is useless in many cases. It is definitely better to show the actual error message from the JSON response. But it seems that the JSON format is different for each service...
> > 
> > 
> > If you look the tagged commit carefully you'll see it was showing proper error message from json data. Also it has different error handling for each service.
> 
> That might be true for some services, but that commit did not remove any error handling from `atlassian/bitbucket`. I just verified that. Without my patch and with the referenced commit reverted you get something like (for a random error that I hit today):
> 
> ```
> requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.bitbucket.org/2.0/repositories/openvpntechnologies/ovpn3-build/commit/4e857d6c9df6956932ed0b25abb8454892dcebea/statuses/build
> ```
> 
> With my patch you get:
> 
> ```
> requests.exceptions.HTTPError: Authenticated user is neither the creator (frank_lichtenheld) of this build status nor an admin of this repository.
> ```
> 
> So I think this is a definite improvement for Bitbucket Cloud specifically.

How about reverting that commit and keeping your patch too?

flichtenheld:
> How about reverting that commit and keeping your patch too?

I don't have a specific opinion on that. But I will definitely not make that part of my PR.

gonchik:
@salmannotkhan  does it really need to be revert it ? 
I dig into code, it was to quite duplication. 
I will be happy to understand the exact usage

salmannotkhan:
> @salmannotkhan does it really need to be revert it ?
> I dig into code, it was to quite duplication.
> I will be happy to understand the exact usage

Okay, let me explain what happened.

I'm using this library for 6 months never faced any issues. But one day I was making a python script to update labels on every bitbucket that I have. 

I didn't know that repo labels can't contain uppercase characters. So I tried to set labels that are uppercase but instead of returning the error from API saying that "Labels can't contain the uppercase letter" I earn a KeyError: "errorMessage" key doesn't exist on d or Something like that.

That means error messages aren't handled properly after the try to handle all errors messages singly because each service has its error message format.

gonchik:
@salmannotkhan noted thank you. My bad in that situation. 
Is it ok if start only for BB cloud, because I use for Jira, Confluence on-prem and error handling works well?
Or I use a little bit in other way then other developers

gonchik:
@salmannotkhan  let me upload a new release. 
Too bad to read once someone meet errors

salmannotkhan:
> @salmannotkhan  let me upload a new release. 
> Too bad to read once someone meet errors 

No problem, sir it's a great library for atlassian services

clrusby:
@gonchik - given it does affect other clients (eg bitbucket server), could we amend your commit 50b5c71 so it reads something like this instead?
```
error_msg = "\n".join(j.get("errorMessages", []) + [k + ": " + v for k, v in j.get("errors", {}).items()])
```

gonchik:
@clrusby sure!

clrusby:
What I wrote doesn’t actually work for bitbucket server as errors is a `list` of `dict`.
It does get a bit messy to handle it generically here, which makes me wonder whether it’s better to add a bespoke error handler for bitbucket server.

PR #932, opened 2022-01-27
[Confluence] Obtaining license information, max users and count

Caramba77:


atlassian/confluence.py
```diff
@@ -2667,6 +2667,34 @@ def clean_jira_metadata_cache(self, global_id):
...
         return self.delete(url, params=params)
 
+    def get_license_details(self):
+        """
+        Returns the license detailed information
+        """
+        url = "rest/license/1.0/license/details"
+        return self.get(url)
+
+    def get_license_usercount(self):
```
gonchik, line 2677: can we have?
```def get_license_user_count(self):```

Caramba77: Sure.  Change committed with new name.

PR #939, opened 2022-02-07
Fix paging when querying Jira via JQL

ilmagico:
It seems that, probably by accident, the `Jira.jql` method doesn't check for paging and, as a result, only returns the first page. This makes it properly retrieve all the pages.

We've been using this patch at my workplace for a while already, but I haven't opened a PR before due to being busy (lazy).
I'm not sure how to run the tests, or how to validate it with "black" as in the contributing instructions, please let me know if I need to fix something.


gonchik:
@ilmagico  Thanks!

PR #943, opened 2022-02-13
Jira: Fix user properties endpoints (use account id)

janoskut:
According to https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-user-properties/#api-group-user-properties, the user identifier for user property endpoints is "accountId", not "username".

This fixes the API for the endpoints:
- Get user property keys
- Get user property
- Set user property
- Delete user property


gonchik:
Looks like we have to support both way, as wrapper used for the on-premises solutions which is mostly for the Server/Data Center releases

PR #949, opened 2022-03-02
[Confluence] page_exists method without error log

Starkandco:
Confluence.py change to page_exists() function. Function will now make its own request and handle the response to determine the True or False value rather than utilising "self.get_page_by_title(space, title)" which outputs errors to the log if it cannot find a page.


Starkandco:
Relates to issue 547

PR #955, opened 2022-03-28
updated get_file_list method for folder support

jaromieee:
To only list files from a specific folder in the target repository, you can add it to the URL like so:
`.../projects/{project_key}/repos/{repository_slug}/files/{folder}

I added an extra parameter to the method, which will add the folder to the URL if given.

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -2200,7 +2200,6 @@ class Bitbucket(BitbucketBase):
             params["limit"] = limit
         return self._get_paged(url, params=params)
 
-
     def get_content_of_file(self, project_key, repository_slug, filename, at=None, markup=None):
         """
         Retrieve the raw content for a file path at a specified revision.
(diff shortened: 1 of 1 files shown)


atlassian/bitbucket/__init__.py
```diff
@@ -2166,20 +2166,23 @@ def create_code_insights_report(
...
         :param repository_slug:
+        :param sub_folder: a sub folder in the target repository to list the files from.
         :param query: the commit ID or ref (e.g. a branch or tag) to list the files at.
                       If not specified the default branch will be used instead.
         :param limit: OPTIONAL
         :return:
         """
         url = "{}/files".format(self._url_repo(project_key, repository_slug))
+        if sub_folder:
+            url = "{}/{}".format(url, sub_folder)
```
abloemert, line 2185: Maybe good to strip the slash at the front if it's given?
```suggestion
            url = "{}/{}".format(url, sub_folder.lstrip("/"))
```

atlassian/bitbucket/__init__.py
```diff
@@ -2189,6 +2200,7 @@ def get_file_list(self, project_key, repository_slug, query=None, start=0, limit
             params["limit"] = limit
         return self._get_paged(url, params=params)
 
+
```
abloemert, line 2203: This line should be removed
```suggestion
```

PR #995, opened 2022-06-10
Confluence: Add page width to page creation

PaulBigPay:
Confluence: Added the ability to create_page in confluence.py to set page full width. Default is False to make it how it is today.


sidr-bigpay (approved):

gonchik:
Noted, thanks!

PR #1000, opened 2022-06-23
Update JIRA docs with missing prefix

Ry-DS:
While this is small, it did confuse me at first and led me to believe this function is in the global namespace.


gonchik:
Thanks ! :)

PR #1018, opened 2022-08-02
[Jira] Fix the parameter of remove_user_from_group from username to account id 

YugoHino:
According to the API [document](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-groups/#api-rest-api-3-group-user-delete), username is no longer available for remove_user_from_group. When this api is called, there is an error "HTTPError: Specified user does not exist or you do not have required permissions".
This PR is fixed the parmeter as following,

username -> accountId

Please let me know if I need to fix something.


gonchik:
@YugoHino  Can we support for the DC release as well ? :) 
it means on prem used the username parameter instead of account_id

YugoHino:
@gonchik Thank you for the comment. Your comment is correct, I only cared about Cloud version but I should care about DC as well. I will change the approach that currently replace username to account_id but username should be there and account_id should be added as optional. I will fix it.

YugoHino:
Hi,

I fixed it. Those params are selected if the url domain is Cloud or not. When the url domain has atlassian.net, accountId is used for params. And else, username is used for it as well as so far.

YugoHino:
Hi @gonchik Sorry for bothring you. Regarding this PR, please let me know if I should do for it. Or, I wait for the review.

gonchik:
Hi @YugoHino , 
let me formalize the existing code and after I will merge it

PR #1022, opened 2022-08-04
Find users who have access to a project to browse

chand912:
Find users who have access to browse issues in a project. Supplying a username=" " will return all the users who have browse permission to a project.
https://docs.atlassian.com/software/jira/docs/api/REST/8.20.11/#user-findUsersWithBrowsePermission

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -785,6 +785,37 @@ class Jira(AtlassianRestAPI):
         params = {"groupname": group_name, "username": username}
 
         return self.delete(url, params=params)
+    
+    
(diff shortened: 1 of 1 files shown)


chand912:
Please review the #1022, very useful to retrieve users who have access to a project (Browse issues level permission), To find all the users who have access to a project by giving empty username (username= " ").

gonchik:
Could you add that method into jira.rst file , like for the doc

chand912:
Created a new pull request (#1024) with new method added to the jira.rst file.

PR #1031, opened 2022-08-16
Add new createmeta functions to comply with Jira 9 API

### LINKED ISSUE #1030: Issue createmeta is outdated and will be deprecated
The current use of the createmeta function in this library is outdated. This article describes how it will be updated:
https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html

We should fix this asap before the endpoint ends up failing for users of this library as they update to Jira 9.

Ry-DS:
This PR closes #1030 

Adds new functions to maintain the same level of functionality, while also adding a deprecation warning for the old function.

To learn more about why this is needed, check out https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html

TODO:
- May be best to use the `self.resource_url` function for consistency's sake

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -1,7 +1,6 @@
 # coding=utf-8
 import logging
 import re
-from typing import Union
 from warnings import warn
 
 from requests import HTTPError
@@ -918,7 +917,7 @@ class Jira(AtlassianRestAPI):
             query_result, missing_issues = self.bulk_issue(issue_list, fields)
         return query_result, missing_issues
 
-    def issue_createmeta(self, project: str, expand="projects.issuetypes.fields"):
+    def issue_createmeta(self, project, expand="projects.issuetypes.fields"):
         """
         This function is deprecated.
         See https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html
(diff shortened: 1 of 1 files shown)


atlassian/jira.py
```diff
@@ -1,6 +1,8 @@
 # coding=utf-8
 import logging
 import re
+from typing import Union
```
gonchik, line 4: does that dependency typing is built-in ?

Ry-DS: Typing is built into python 3, shouldn't be an issue if it's included.

gonchik: But we have still python2 clients ;(

Ry-DS: you should add python2 to the ci, as right now it only contains python 3. Rolled back the typing changes here: 84d5bc5  :+1:

gonchik: Main idea is remove python2 support, but time to time I receive emails ;)
That's why I have mentioned ;)

pmilosev:
@Ry-DS please see my comment here: https://github.com/pycontribs/jira/pull/1527#issuecomment-1315931769

It looks like cloud Jira and Jira DC/Server have independent REST API and any client connecting to it should account for potential differences. Cloud JIRA still uses the `createmeta` endpoint and does not support the new API.
This probably doesn't affect your change here as you only added functions that use the new endpoints, but you might want to check your deprecation warning and eventually guard it with some `if` statements as the `createmeta` endpoint is not deprecated in general.

(you don't need to act on this, but since I used your PR as reference on how to implement the support in `pycontrib/jira` I thought sharing this info)

gonchik:
@pmilosev  thanks for that mention. 
I see interesting conditions related to the version checkers.

PR #1064, opened 2022-10-25
Jira: Add functionality to get comments on an issue

martynbristow:
Implement:

 - issue_get_comments(issue_id)
 - issue_get_comment(issue_id, comment_id)
 - issues_get_comments_by_id(comment_id, [comment_id ...])

### DIFF
--- a/examples/jira/jira_get_comment.py
+++ b/examples/jira/jira_get_comment.py
@@ -1,8 +1,7 @@
 # coding=utf-8
 from atlassian import Jira
 
-""" Get Comments on an Issue
-"""
+"""Example: Get Comments on a Jira Issue"""
 
 
 def main():
--- a/examples/jira/jira_get_comment.py
+++ b/examples/jira/jira_get_comment.py
@@ -1,7 +1,7 @@
+""" Example: Get Comments on a Jira Issue
+"""
 # coding=utf-8
 from atlassian import Jira
-
-"""Example: Get Comments on a Jira Issue"""
 
 
 def main():
--- a/examples/jira/jira_get_comment.py
+++ b/examples/jira/jira_get_comment.py
@@ -1,0 +1,25 @@
+# coding=utf-8
+from atlassian import Jira
+
+""" Get Comments on an Issue
+"""
+
+
+def main():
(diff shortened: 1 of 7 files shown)


isaac-philip:
I can have a look at this.

examples/jira/jira_get_comment.py
```diff
@@ -0,0 +1,24 @@
...
+    print("\n*Get Comments on INT-1*")
+    comments = jira.issue_get_comments("INT-1")
+    print(comments)
+
+    print("\n*Get a single comment on INT-1*")
+    comments = jira.issue_get_comment(10000, 10001)
+    print(comments)
+
+    print("\n*Get a set of comments*")
+    comments = jira.issues_get_comments_by_id(10000, 10002)
```
isaac-philip, line 19: just a suggestion for the naming of function,
keeping in line with the single comment function naming convention,
could we keep multiple `comments` for the single `issue` named as,
`issue_get_comments_selective` rather than `issues_get_comments_by_id`
to help us understand that it is multiple selected comments for single issue.
Or some other better name? because i believe the filtering would always be based on the comment_id so keeping by_id mite be redundant here.

because it would always be comments drawn for the single issue and not multiple one if i understand correctly.

isaac-philip (changes requested):
Great work! looks good, just a small name change if we can have it, thanks

examples/jira/jira_get_comment.py
martynbristow: Yeah, I'm happy to change it to this. My only reason for choosing: `issues_get_comments_by_id` was to be consistent with the Atlassian JIRA API docs: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-comments/#api-rest-api-3-comment-list-post

`issues_get_comments_by_ids is still a bit weak`, so I'll consider selective unless I have an idea

I also chose: `issues_get_comments` rather than `get_issue_comments` to match some existing code

isaac-philip: that sounds fine then, looks good to me for merge, thanks!

isaac-philip:
Please have final review and merge if good @gonchik

PR #1066, opened 2022-10-27
Fix print typo in code block example in README

MatthewJohn:


gonchik:
@martyn-bristow Let's merge it. 
I think we have to improve our docs :)

gonchik:
@martyn-bristow hm, you know you're totally right. 
I made a small mistake :(

MatthewJohn:
Sorry @martyn-bristow, I was skimming too much :( Sorry for causing such a palaver!

PR #1074, opened 2022-11-08
Resolve Failing Tests: Correction to adanced_usage=True

martynbristow:
`return None` instead of response


martynbristow:
@gonchik ... I've made a change to make the tests pass while preserving functionality

Be interested in your perspective

gonchik:
Hi @martynbristow  
I do support the exact return. 

Thanks for the PR

martynbristow:
The build should go green now
Hope that's all good now

PR #1079, opened 2022-11-15
Added request_current_index_from_node method and docs

SLRover:
Added request_current_index_from_node method and docs for the cluster methods


gonchik:
Thanks for the PR. 
looks like black checker is needed

Issue #1057, opened 2022-10-12
[Confluence Cloud] get_page_as_pdf  is not working


rootsj:
log.error PDF conversion not successful

I used it well until October 6th,
but it's not working since October 7th.

glenlobo:
Facing the same issue.
Throws a message "PDF conversion not successful."

Zad2:
Me too, do you think that's related to https://developer.atlassian.com/cloud/confluence/changelog/#CHANGE-720 ?

rootsj:
Yes, I think it's related.
I'm using it after customizing it temporarily.

Zad2:
@rootsj can you share how you customized it to fix it?

rootsj:
I modified def get_pdf_download_url_for_confluence_cloud of confluence.py.

`            long_running_task = True
            headers = self.form_token_headers
            log.info("Initiate PDF export from Confluence Cloud")
            response = self.get(url, headers=headers, not_json_response=True)
            response_string = response.decode(encoding="utf-8", errors="strict")
            task_id = response_string.split('name="ajs-taskId" content="')[1].split('">')[0]
            poll_url = "/services/api/v1/task/{0}/progress".format(task_id)
            while long_running_task:
                long_running_task_response = confluence.get(poll_url, headers=headers, not_json_response=True)
                long_running_task_response_parts = json.loads(long_running_task_response.decode(
                    encoding="utf-8", errors="strict"
                ))
                percentage_complete = long_running_task_response_parts["progress"]
                is_update_final = long_running_task_response_parts["progress"] == 100
                current_state = long_running_task_response_parts["state"]
                log.info("Sleep for 5s.")
                time.sleep(5)
                log.info("Check if export task has completed.")
                if is_update_final and current_state == "UPLOADED_TO_S3":
                    log.info(percentage_complete)
                    log.info("Downloading content...")
                    log.debug("Extract taskId and download PDF.")
                    download_url = long_running_task_response_parts["result"][6:]
                    long_running_task = False
                elif not is_update_final and current_state == "FAILED":
                    log.error("PDF conversion not successful.")
                    return None
                else:
                    log.info(percentage_complete)`

using def get_page_as_pdf
Then, you can know s3 url for pdf download.

`
s3_url = confluence.get_page_as_pdf(page_id)
response = requests.get(s3_url)
pdf_content = reponse.content
`

If you don't want to modify the lib,
refer to the code and make it separately

[...]

PR #1081, opened 2022-11-21
[confluence] Fix def get_page_as_pdf (pdf Export was rollback)

rootsj:
https://developer.atlassian.com/cloud/confluence/changelog/#CHANGE-754

> Note: As these endpoints are currently experimental, they can change without any prior deprecation notice. Please be aware of this before using them in business critical applications.

If you check the change-754,
As Note, S3 url is considered a temporary end point.
So Since 11/18,
I checked the get_page_as_pdf, is not working.
It seems that the structure has been changed to the previous structure.

checking the confluence source,
it works with previous code.
So, PR with Revert commit.

Revert "[confluence] Fix def get_page_as_pdf When using Confluence Cloud(#1057)"

This reverts commit 04309dbaebbc95d66e77c247ed02cf24a9218ffa.

* Since November 18th, 2022
* Confluence Cloud PDF Export was rollback

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -4,7 +4,7 @@ import os
 import time
 import json
 
-from requests import HTTPError, get
+from requests import HTTPError
 from deprecated import deprecated
 from atlassian import utils
 from .errors import (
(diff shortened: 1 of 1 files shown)


gonchik:
Thanks!

PR #1087, opened 2022-11-27
Python 2.7 compatibility

orent:
As long as Python 2.7 support is maintained, it should at least be importable without  SyntaxErrors, even if full functional testing is not part of the CI.


gonchik:
Thank you for your PR :)

PR #1088, opened 2022-11-28
Ooops. Fixed Python 2.7 import

orent:


gonchik:
@orent Thanks for that fix

PR #1090, opened 2022-11-29
Returned _link details for next loop

tpatel-leena:
To get all the all the pages need to pass all the details related to next traversal

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -511,7 +511,7 @@ class Confluence(AtlassianRestAPI):
 
             raise
 
-        return response.get("results")
+        return response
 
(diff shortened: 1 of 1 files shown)


gonchik:
@tpatel-leena  how about to make a call like, 
```while get_all_pages_from_space():```

Because I see many scripts have writter to old method response,

tpatel-leena:
@gonchik

My PR was merged 12 days back
And is working flawlessly, I tested all the methods

tpatel-leena:
@gonchik 
How will I know whats the next page if there are only results ?
We would need all the next details to figure out that there is a next page

PR #1091, opened 2022-11-30
Add the type parameters whether a page exists

renwofei423:
Add the type parameters when judging whether a page exists, such as blogpost, page


renwofei423:
without this change, you can't judge a blog whether exists.

gonchik:
Thank you for the PR @renwofei423

PR #1106, opened 2023-01-05
Add custom field editing

PaulBigPay:
Added some add an remove commands for custom fields in Jira


PaulBigPay:
@gonchik Not sure why Black is failing 

black --check --diff atlassian/ examples/ tests/ --exclude __pycache__ 
All done! ✨ 🍰 ✨
206 files would be left unchanged.

works fine on local, any ideas?

PR #1119, opened 2023-02-09
Fix API url for getting objecttype attributes

aohontsev:
Hi!
I get an error when trying to get objecttype attributes via this lib.
The was occurred because of invalid link generation - just miss a slash in generated url:
https://jira.insight.url/rest/insight/1.0objecttype/1/attributes'


gonchik:
@aohontsev Thank you for the detecting

PR #1121, opened 2023-02-15
Jira ServiceDesk: Fix create attachment

### LINKED ISSUE #1056: [Service Desk] Creating/Adding attachments
I found a small issue when trying to add an attachment to an active request:
When calling the method create_attachment from the ServiceDesk class, the following error is raised:
File "/home/thomas/git/private/atlassian-python-api/atlassian/service_desk.py", line 497, in attach_temporary_file
    .json()
AttributeError: 'dict' object has no attribute 'json'.

The error is based on the fact that when using the ServiceDesk class and subsequently the method self.post(...) in normal mode, the response is already converted to JSON inside that method (post). Using the ServiceDesk class in advanced mode does not raise any issues.

skropf:
In service_desk.py/attach_temporary_file a check was added to successfully post the request. fixes #1056

If advanced mode is set the raw response must be converted to json, not so in normal mode.

### DIFF
--- a/atlassian/service_desk.py
+++ b/atlassian/service_desk.py
@@ -79,7 +79,9 @@ class ServiceDesk(AtlassianRestAPI):
 
     def get_my_customer_requests(self):
         """Returning requests where you are the assignee"""
-        response = self.get("rest/servicedeskapi/request", headers=self.experimental_headers)
+        response = self.get(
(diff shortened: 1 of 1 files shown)


gonchik:
@skropf Thank you!

paulafernanda156:
nice report.

Issue #1125, opened 2023-02-22
[Confluence] Unable to pass attachment stored in memory to add_attachment method


dizoncruz:
In the official Python library for JIRA docs, JIRA's add_attachment method expects an issue_key and attachment.  It only expects a filename if the attachment is being passed from memory as a StringIO object .
https://jira.readthedocs.io/examples.html#:~:text=.name)-,2.1.11.%20Attachments,-%23

In the Atlassian Python API documentation, the add_attachment method only expects an issue_key and a filename.  This prevents the attachment of files stored only in memory.  
https://atlassian-python-api.readthedocs.io/jira.html

gonchik:
@dizoncruz we can implement it.

dizoncruz:
@gonchik Perfect, thank you!

dizoncruz:
Thank you for implementing this so quickly.  I see that it's been merged to master, but I'm unable to use the new method.

Using apache-airflow-providers-atlassian-jira, which according to the docs utilizes this Atlassian JIRA SDK
https://airflow.apache.org/docs/apache-airflow-providers-atlassian-jira/stable/_api/airflow/providers/atlassian/jira/operators/jira/index.html

Code:
```
from airflow.providers.atlassian.jira.hooks.jira import JiraHook
hook = JiraHook('jira_web_staging')
jira = hook.client

getattr(jira,'add_attachment_object')(issue['key'],attachment=attachment)
```
Log output:
```
<atlassian.jira.Jira object at 0x7f6a28b3b1f0> <class 'atlassian.jira.Jira'>
AttributeError: 'Jira' object has no attribute 'add_attachment_object'
```

gonchik:
@dizoncruz  I did not released yet that. 
Let me collect changes into 1 chunk and make a release.

dizoncruz:
Hi @gonchik, thank you so much for implementing this.  I can confirm that the add_attachment_object method is working as intended.

One last question: the original add_attachment method takes in an optional filename string in cases where the attachment is a StringIO object in memory.  It looks like this new add_attachment_object does not though.  Would that functionality have to be built in as well, or is there another way I pass in a filename string so that the attached file isn't just called "file"?

gonchik:
@dizoncruz  could you send a PR, i would like to see how do you want to implement it.

PR #1126, opened 2023-02-23
Update jira.py add_attachment method

laralight:
added logic to add_attachment function

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -1191,7 +1191,7 @@ class Jira(AtlassianRestAPI):
             return self.get(url)
         return (self.get(url) or {}).get("fields").get("labels")
 
-    def add_attachment(self, issue_key, filename):
+    def add_attachment(self, issue_key, attachment):
         """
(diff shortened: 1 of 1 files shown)


gonchik:
Thank you for your PR. 
I think it's better to have dedicated method like add_attachment_object()
`    def add_attachment_object(self, issue_key, attachment):
        """
        Add attachment to Issue
        :param issue_key: str
        log.warning("Adding attachment...")
        base_url = self.resource_url("issue")
        url = "{base_url}/{issue_key}/attachments".format(base_url=base_url, issue_key=issue_key)
        if attachment:
            files = {"file": attachment}
            return self.post(url, headers=self.no_check_headers, files=files)
`

and rewrite previous method into the next one: 
`
  def add_attachment(self, issue_key, filename):
        with open(filename, "rb") as attachment:
               return self.add_attachment_object(issue_key, attachment)
`

gonchik:
that's exact the requested method  #1125

PR #1135, opened 2023-03-17
This will give you a fixed-width page.

zackn9ne:
This will give you a fixed-width page.


zackn9ne:
At Confluence, for `update-page` method, if you already have a `full-width` page and you send `full-width: False` the logic does not work. You still end up with a `full-width` page.

This patch fixes that.

PR #1160, opened 2023-05-04
Add the possiblity to change the signature method

h-braun:
RSA_SHA512 was hardcoded in the REST Client for OAuth1. It is now possible to change the 'Signature Method' in the 'oauth_dict' with the key 'signature_method' to another signature. The default value is kept as 'SIGNATURE_RSA' if no value for the mentioned key is provided.


gonchik:
@h-braun  Thank you for your contribution.

Issue #1158, opened 2023-05-03
PDF export function no longer working


spadgezilla:
Hi, I'm having an issue with the PDF export function, this was working up until May 1st, I wonder if recent changes found here could be to blame? https://confluence.atlassian.com/cloud/blog/2023/05/atlassian-cloud-changes-apr-24-to-may-1-2023

I'm using confluence.export_page(page_id) and getting 'PDF conversion not successful' in return. It also seems to return the html content of the page in IDLE shell so it's doing something, I can't remember if it used to do this though.

Is anyone else having this problem?

rosomri:
Same

rosomri:
@torrid @spadgezilla 
Hi there! 
I've just created a PR that aims to solve this issue.
It seems to be working fine for me, would appreciate your feedback.

**Here's the current flow works:**
1.  Create a pdf page export task using the "spaces/flyingpdf/pdfpageexport.action?pageId={pageId}" endpoint.
2. In the next step, use the poll endpoint, "runningtaskxml.action?taskId=[taskId]", to determine whether the export task has been completed.
3. Retrieve the download URL from the poll endpoint response if the task was successfully completed
4. Get the PDF data from the retrieved endpoint.

**Step 2 seems to be where the issue appears:**
- It does not matter if the task is successfully initialised or not, the poll endpoint will report that the task does not exist without raising an error message.
- As a result, step 3 queries the space URL without a specific page.
- Consequently, the response is a general HTML page without any error messages.

**The problems are as follows:**
- There seems to be a problem with the poll endpoint. Atlassian recently updated Confluence's PDF export functionality.
- An error should be thrown if the task ID is invalid.

**Optional solution:**
**TLDR;**
**The functionality is working great, but the PR has not yet been completed.**

- Atlassian recently updated Confluence's PDF export functionality. To get the export task's current state, `task/{taskId}/progress` endpoint should be used instead of the deprecated `runningtaskxml.action?taskId={taskId}`. 
- Also the flow of retrieving the content from the download signed URL was changed a bit.
In order to download a PDF file from a signed S3 URL, you do not need to provide any authentication or headers. If you do provide those, the request will fail. As the session has already been initiated with those headers, it cannot be used.

**Feel free to edit and elaborate**

pp-wdiebel:
I am running into this error as well, and this appears to be what is going on in the python:

[...]

PR #1161, opened 2023-05-04
[Confluence]: PDF export function no longer working #1158 BUGFIX

rosomri:
## #1158 bugix
TLDR; 
Atlassian recently updated Confluence's PDF export functionality.

**Here's the current flow works:**
1.  Create a pdf page export task using the `spaces/flyingpdf/pdfpageexport.action?pageId={pageId}` endpoint.
2. In the next step, use the poll endpoint, `runningtaskxml.action?taskId={taskId}`, to determine whether the export task has been completed.
3. Retrieve the download URL from the poll endpoint response if the task was successfully completed
4. Get the PDF data from the retrieved endpoint.

**Step 2 seems to be where the issue appears:**
- It does not matter if the task is successfully initialised or not, the poll endpoint will report that the task does not exist without raising an error message.
- As a result, step 3 queries the space URL without a specific page.
- Consequently, the response is a general HTML page without any error messages.

**The problems are as follows:**
- There seems to be a problem with the poll endpoint. Atlassian recently updated Confluence's PDF export functionality.
- An error should be thrown if the task ID is invalid.

**PR main changes**
**The functionality is working as expected**
- Atlassian recently updated Confluence's PDF export functionality. To get the export task's current state, `task/{taskId}/progress` endpoint should be used instead of the deprecated `runningtaskxml.action?taskId={taskId}`. 
- Also the flow of retrieving the content from the download signed URL was changed a bit.
In order to download a PDF file from a signed S3 URL, you do not need to provide any authentication or headers. If you do provide those, the request will fail. As the session has already been initiated with those headers, it cannot be used. Does it make sense to use the request library directly?
- **Feel free to edit and elaborate**

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -2438,7 +2438,7 @@ class Confluence(AtlassianRestAPI):
                 log.error("Failed to get download PDF url.")
                 raise ApiNotFoundError("Failed to export page as PDF", reason="Failed to get download PDF url.")
             # To download the PDF file, the request should be with no headers of authentications.
-            return requests.get(url).content
+            return requests.get(url, timeout=75).content
         return self.get(url, headers=headers, not_json_response=True)
 
     def get_page_as_word(self, page_id):
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -5,6 +5,7 @@ import time
 import json
 
 from requests import HTTPError
+import requests
 from deprecated import deprecated
 from atlassian import utils
 from .errors import ApiError, ApiNotFoundError, ApiPermissionError, ApiValueError, ApiConflictError, ApiNotAcceptable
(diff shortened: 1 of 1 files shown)


cforce:
Please merge. It blocks usage of the Confluence apis:
I try to use it based on direct import but still struggle - see https://github.com/operatorequals/httpimport/issues/53

atlassian/confluence.py
```diff
@@ -2669,45 +2674,40 @@ def get_pdf_download_url_for_confluence_cloud(self, url):
...
-                        current_status = long_running_task_response_parts[3]
-                        download_url = current_status.split("href=&quot;/wiki/")[1].split("&quot")[0]
-                        long_running_task = False
-                    elif is_successful == "<isSuccessful>false</isSuccessful>":
+                progress_response = self.get(poll_url)
+                percentage_complete = int(progress_response.get("progress", 0))
+                task_state = progress_response.get("state")
+                if percentage_complete == 100:
+                    running_task = False
+                    log.info(f"Task completed - {task_state}")
```
gonchik, line 2692: no f-string please

cforce:
Please release

cforce:
This bugfix is still not released -what is preventing it?

PR #1188, opened 2023-06-18
Make check_inbox_pull_requests to return paged generator

noorul:
Make check_inbox_pull_requests to return paged generator


gonchik:
@noorul  Thanks

PR #1205, opened 2023-07-24
Use more performant version of Python logging calls

madscientist:


madscientist:
Basically, the logging string is not formatted until/unless the log message is going to be emitted.  So for, for example, debug() messages which are rarely if ever actually emitted to a log sink, this means Python doesn't have to allocate the string and format it all.  See for info https://docs.python.org/3/library/logging.html#logging.debug

If you use .format() then the formatted string is constructed by the calling function before the logger even gets to see it, so it cannot avoid the formatting even if the resulting formatted string is just dropped on the floor.

gonchik:
Thanks for explanation :) @madscientist

PR #1207, opened 2023-07-27
[Jira] Commits parameter addition for until and since

thejosess:
This PR adds two parameter (until, since) on get_commits which is only available for Jira Server.

Thank you.


gonchik:
@thejosess thanks

PR #1208, opened 2023-07-28
Pipeline triggering examples with a request with pipeline variables

m0nt3cr1st0:
Hey guys! I was using the API for on of my projects and I had to browse the source code in order to understand how to pass variables which is not optimal, so I though an example of that would help!


gonchik:
Thanks! let me merge it

PR #1215, opened 2023-08-15
Fix packaging of the tests directory

### LINKED ISSUE #1206: tests/mockup.py missing from the PyPi source tarball
The [latest tarball](https://files.pythonhosted.org/packages/b8/75/6090a372e9b1a9221f67be256e517c9e663cfd6bd5426815eb7de03f32b5/atlassian-python-api-3.39.0.tar.gz) doesn't contain the file, without which it is impossible to run the tests.

WOnder93:
Do not install tests as a package, but instead include it only in the source tarball via MANIFEST.in.

I have verified that the resulting tarball installs only the 'atlassian' package into site-packages and that the included files are sufficient to successfully run the tests.

Fixes #1206


gonchik:
@WOnder93  thanks

darthwalsh:
Thanks for this fix! We were seeing `ModuleNotFoundError: No module named 'tests.helpers'` in our pytest project until we installed a newer version.

PR #1224, opened 2023-08-18
Bamboo: Added 2 methods to fetch deployment projects associated with …

gonchik:
…a plan and to trigger a deployment of a release to an env.


gonchik:
was merged

PR #1237, opened 2023-08-31
Documentation: Bitbucket project_grant_group_permissions parameter

elliot-martin-a:
The documentation lists "groupname" as the parameter accepted by project_grant_group_permissions, but group_name is expected:
https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/bitbucket/__init__.py#L500


gonchik:
thanks for the detection :)

PR #1239, opened 2023-09-01
Jira: updated worklog urls

Bertsjuhn:
Now using the 'resource_url' function to work with 'api_version' configuration.
Added a new function to get a list of deleted worklogs.


gonchik:
Thanks

Issue #1241, opened 2023-09-09
advanced_mode flag is not passed to underlying `request` method preventing handling of 429 rate limit


2ps:
several methods of `atlassian.rest_client.AtlassianRestAPI`, like the `post` method, do not propagate the `advanced_mode` flag to the underlying `request` method.  Because that flag is not propagated to the underlying call, rate limit exceptions are thrown and graceful handling of a 429 status code by using the response headers is not possible.  

Proposed fix:  pass the `advanced_mode` flag passed into functions like `post` to the underlying `request` method.

gonchik:
@2ps please share PR

Spacetown:
It will be better to handle the 429 on our own, hiding the logic from the user. The user never wants to abort the script because of rate limits.

We should work on #904.

2ps:
This can be closed, it is addressed in a merged PR.

PR #1243, opened 2023-09-09
[ISSUE-1241] #1241, added propagation of `advanced_mode` flag 

2ps:
#1241 

added propagation of the `advanced_mode` flag for `put`, `post`, `patch`, and `delete` functions.  Prior to this change, the `advanced_mode` flag was only propagated to the `request` method for `get` calls.  The `advanced_mode` `requests.Response` object is very useful in having code gracefully handle rate limit 429 responses from the server since the response headers will include the necessary backoff information that the client should use before making another request.

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -343,6 +343,7 @@ class AtlassianRestAPI(object):
             params=params,
             trailing=trailing,
             absolute=absolute,
+            advanced_mode=advanced_mode,
         )
         if self.advanced_mode or advanced_mode:
             return response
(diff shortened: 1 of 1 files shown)


gonchik:
@2ps  could you update a docs please?

2ps:
> @2ps could you update a docs please?

Apologies for missing this.  Will address in a separate PR since this one has already been merged.

2ps:
> @2ps could you update a docs please?

My bad, it looks like this was already included in the docs for put, post, patch, and delete but just had not been wired in.  At least from my review there were no docs to update.  Please do point me in the right direction if I have overlooked something.

Issue #1242, opened 2023-09-09
rest_client.py error on line 481 when values or keys are integers


2ps:
We should use `%`, `.format`, or f-strings instead for situations when keys or values are integers, since `+` does not automatically type-convert for us.

Spacetown:
I think that the user should provide the data with the correct types. There can be parameters which need to be integers.

2ps:
This is in exception handling, I don't see how the user can specify which values are in the response header, although I am happy to hear suggestions.

2ps:
[https://github.com/atlassian-api/atlassian-python-api/pull/1244](url) addresses this issue.

2ps:
@Spacetown  @gonchik  could I please get your review on this at your convenience?

Spacetown:
Ok, for exception handling this can be done. Next time you should use a permalink instead of a line number without a link.

Zircoz:
Fixed in `rest_client.py`: URL parameters are now encoded via `urllib.parse.urlencode()` (line 466), which correctly handles integer keys and values — no more `TypeError` from string concatenation.

PR #1244, opened 2023-09-09
[ISSUE-1242]  prevented weird exceptions related to integer handling

2ps:
if keys or values in the response json are `int`s, the `k + ": " + v` expression will fail because an `int` is not a `str`.  In order to avoid this, we can either use `%`-formatting, `.format` formatting, or f-strings.  I chose f-strings for usability, but regardless of which direction, the issue should be addressed one of those ways.

addresses #1242

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -477,7 +477,7 @@ class AtlassianRestAPI(object):
             try:
                 j = response.json()
                 if self.url == "https://api.atlassian.com":
-                    error_msg = "\n".join([k + ": " + v for k, v in j.items()])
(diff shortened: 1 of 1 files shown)


gonchik:
@2ps  could wrap via str() instead of f-strings. 
Because it will break 2.7 compatibility

gonchik:
@2ps please use format in that situation please?

2ps:
Apologies

> @2ps please use format in that situation please?

Apologies for the late response, will do.  That said, the setup.py file indicates that python2 isn't supported -- we may also want to update that to indicate support for python 2 and the eol'd python 2.7.

gonchik:
Thanks

PR #1245, opened 2023-09-13
feat(insight): implement get objectschema objecttypes flat

gardar:


gonchik:
@gardar thank you! :) 
also, I will be happy if you add some word into insight.rst file ;)

gardar:
> also, I will be happy if you add some word into insight.rst file ;) 

Sure, but where is it located?

gonchik:
@gardar it's not generating automatically :(

gardar:
> @gardar it's not generating automatically :( 
> 

I would think that would be trivial to add, since you are already using sphinx and have docstrings for most functions, you even have the autodoc extension loaded 🙂

https://www.sphinx-doc.org/en/master/tutorial/automatic-doc-generation.html

gonchik:
@gardar  will be happy if you fix that :)

PR #1255, opened 2023-09-28
Add Content-Type to calls to Assets in cloud

tester22:
I experienced 415 error messages due to missing headers when making put and post calls to Assets in cloud.


gonchik:
@tester22 Thanks!

PR #1262, opened 2023-10-16
Add Python 3.12 CI build

offa:
Initial steps to support Python 3.12.


offa:
Wired, setuptools seems missing even though it's installed.

```
#22 1.434       ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
```

gonchik:
https://github.com/pypa/setuptools/issues/3661

PR #1326, opened 2024-02-03
Fixed Pull Requests Activities path

roboqo:
Non existing API endpoint was used in the latest version, so this PR fixes that.

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -1781,7 +1781,7 @@ class Bitbucket(BitbucketBase):
         :param start:
         :return:
         """
-        url = "{}/activities".format(self._url_pull_request(project_key, repository_slug, pull_request_id))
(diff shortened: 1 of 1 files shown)


SkylerSalvato:
@roboqo @gonchik 

Hi, my team uses this with the Bitbucket Server REST API, and the endpoint actually still uses `activities`.
Would it be possible to implement this change specifically for Bitbucket Cloud, similar to the following?

https://github.com/roboqo/atlassian-python-api/blob/94d8475a445a5890258ff3049c149a8a7faa7ca4/atlassian/bitbucket/__init__.py#L1552-L1556

SkylerSalvato:
See my proposed change here: #1336

taniajtaniaj:
Hi,
This change breaks our system. Our bitbucket supports "activities" but not "activity" ...

gonchik:
@roboqo  @taniajtaniaj I will publish a new release in a few hours

taniajtaniaj:
thanks a lot! very much appreciated!!

PR #1339, opened 2024-02-26
[Rest] Support exponential backoff and retry

flichtenheld:
Uses the urllib3 builtin Retry mechanism.
We only support retrying on HTTP return status,
not on any of the other retries that that class
defines (e.g. connection problems, redirects).

The usual use case for this is to handle 429
responses from the Atlassian Cloud API which
has request limits that are easily hit by
scripts iterating over objects like Jira
issues or Bitbucket pull requests.

Based on earlier work by @jp-harvey.

Replaces #904

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -3,6 +3,7 @@ import logging
 from json import dumps
 
 import requests
+from requests.adapters import HTTPAdapter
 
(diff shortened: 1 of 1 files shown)


gonchik:
@flichtenheld Thanks!

PR #1342, opened 2024-03-01
rest client: Only enable `backoff_and_retry` if `urllib3` is at least 2.0.0

djgoku:
We are trying to use `backoff_and_retry` but are getting errors with urllib3 1.26.X this fixes and sets the minimum of `urllib3` to `>= 2.0.0`

The first release supporting `backoff_jitter` in the `Retry` class is the `2.0.0` release.

https://github.com/urllib3/urllib3/releases/tag/2.0.0

It doesn't look like this will be added to `1.26.X` https://github.com/urllib3/urllib3/pull/2952#issuecomment-1523888273

### DIFF
--- a/atlassian/rest_client.py
+++ b/atlassian/rest_client.py
@@ -13,6 +13,7 @@ from requests import HTTPError
 from requests_oauthlib import OAuth1, OAuth2
 from six.moves.urllib.parse import urlencode
 from urllib3.util import Retry
+import urllib3
 
 from atlassian.request_utils import get_default_logger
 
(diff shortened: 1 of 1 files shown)


timsavage:
This is problematic for companies using RHEL.  urllib3>2.0 requires a version of libssl that is not available on older (but still supported releases) or RHEL.

djgoku:
> This is problematic for companies using RHEL.  urllib3>2.0 requires a version of libssl that is not available on older (but still supported releases) or RHEL.

Ok, I'll look into alternative options. Thanks.

djgoku:
> This is problematic for companies using RHEL. urllib3>2.0 requires a version of libssl that is not available on older (but still supported releases) or RHEL.

Here is another option I thought of. Hacky, but at least it won't fail is `urllib3` is less than version 2.

djgoku:
@gonchik can I get a review on this?

PR #1343, opened 2024-03-02
[Jira] fix paginated projects retrieving

### LINKED ISSUE #1310: endless loop when retrieving projects from the cloud
Hi,
we have a pretty large list of projects, and when we retrieve the project list from the cloud, and the list is paginated, the library gets stuck in a loop.
It looks like [projects["nextPage"]](https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/jira.py#L2332 ) is still the same at every iteration, so the url to download the list from is always the same: `https://company.atlassian.net/rest/api/2/project/search?maxResults=50&startAt=50`

I was expecting a `startAt=100`, then `startAt=150`, but it's always 50.

Am I missing something? I am not a Python expert, I just observed this strange behaviour

Thanks!

ebaschiera:
Fixes #1310

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -2328,10 +2328,10 @@ class Jira(AtlassianRestAPI):
         next_page_url = projects.get("nextPage")
         while not is_last_page:
             next_page_projects = self.paginated_projects(
-                    included_archived=included_archived,
-                    expand=expand,
-                    url=next_page_url,
-                )
+                included_archived=included_archived,
+                expand=expand,
+                url=next_page_url,
+            )
             next_page_url = next_page_projects.get("nextPage")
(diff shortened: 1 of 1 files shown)


Spacetown:
This looks like a refactoring only. The logic is still the same.

Spacetown:
@gonchik Are you really sure that this fixes the issue? The loop is only restructured using variables but is still doing the same thing.

gonchik:
@Spacetown hm, it's automatically closed that issue. 
I just merged iterator.

Spacetown:
Yes, the iterative was only refactored. In the closed issue I linked a Branch where the same workaround as for the cloud is added. I raise a PR and you can check it.

PR #1344, opened 2024-03-06
[Jira] Fix paged projects

Spacetown:
Add a generic method for paging in the cloud.

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -24,6 +24,70 @@ class Jira(AtlassianRestAPI):
 
         super(Jira, self).__init__(url, *args, **kwargs)
 
+    def _get_paged(
+        self,
+        url,
+        params=None,
+        data=None,
+        flags=None,
+        trailing=None,
+        absolute=False,
+        paging_workaround=False,
+    ):
+        """
+        Used to get the paged data
+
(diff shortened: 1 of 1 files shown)


atlassian/jira.py
```diff
@@ -24,6 +24,62 @@ def __init__(self, url, *args, **kwargs):
...
+        :param params: dict (default is None):     The parameter's
+        :param data: dict (default is None):       The data
+        :param flags: string[] (default is None):  The flags
+        :param trailing: bool (default is None):   If True, a trailing slash is added to the url
+        :param absolute: bool (default is False):  If True, the url is used absolute and not relative to the root
+
+        :return: A generator object for the data elements
+        """
+
+        if self.cloud:
```
gonchik, line 49: for DC the same logic, but not for all end points. 
how about to add as parameter, like supported method or not.

gonchik: My point was about thinking a generic solution for pages methods

Spacetown: I was also thinking about this but then we need to provide a code for this depending on the module:
https://github.com/atlassian-api/atlassian-python-api/blob/978fa3fcd9265a0f9bdf9d690bf01e220c23de4b/atlassian/jira.py#L69-L77

PR #1356, opened 2024-03-21
[BitBucket] added new method get_groups + docs + examples

sbilque:
added a new method to the Bitbuckets get_groups which:

* Takes a  display name for the group as input.
* Returns a list of user groups.


sbilque:
Checks have failed due to a problem with building of docker images. The proposed code is not to blame.

PR #1400, opened 2024-05-24
Add `Crowd` a new attr

ChowRex:
Add a new property of class `Crowd`, can retrieves full details of all group memberships, see: https://docs.atlassian.com/atlassian-crowd/5.3.1/REST/#usermanagement/1/group-getAllMemberships


gonchik:
@ChowRex  could you wrap up without lxml please? 
like with bs4 ?

PR #1418, opened 2024-06-12
Adding Typehints for jira.py

bandophahita:
Since adding typehints to the entire repo would make for a VERY large PR, I'm breaking up the changes mostly by module.  This PR focuses on adding hints to jira.py

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -1,12 +1,14 @@
 # coding=utf-8
 import logging
 from enum import Enum
+from typing import Optional
 
 from deprecated import deprecated
 from requests import HTTPError
 
-from .base import BitbucketBase
 from atlassian.bitbucket.cloud import Cloud
+
+from .base import BitbucketBase
 
 log = logging.getLogger(__name__)
 
@@ -1891,7 +1893,7 @@ class Bitbucket(BitbucketBase):
 
     def assign_pull_request_participant_role(
         self, project_key: str, repository_slug: str, pull_request_id: int, role: str, user: str
-    ) -> dict:
+    ) -> Optional[dict]:
         """
         Assign a role to a user for a pull request
         :param project_key: The project key
@@ -2117,7 +2119,9 @@ class Bitbucket(BitbucketBase):
         data = {"text": text, "severity": severity}
         return self.post(url, data=data)
 
-    def search_pull_request_blocker_comment(self, project_key: str, repository_slug: str, pull_request_id: int) -> dict:
+    def search_pull_request_blocker_comment(
+        self, project_key: str, repository_slug: str, pull_request_id: int
+    ) -> Optional[dict]:
         """
         Get all comments that block the merge of a pull request
         :param project_key: The project key
--- a/atlassian/bitbucket/cloud/repositories/repositoryVariables.py
+++ b/atlassian/bitbucket/cloud/repositories/repositoryVariables.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-
+from __future__ import annotations
 from ..base import BitbucketCloudBase
 
 
@@ -7,7 +7,7 @@ class RepositoryVariables(BitbucketCloudBase):
     def __init__(self, url, *args, **kwargs):
         super(RepositoryVariables, self).__init__(url, *args, **kwargs)
 
-    def __get_object(self, data):
+    def __get_object(self, data) -> RepositoryVariable:
         return RepositoryVariable(
             self.url_joiner(self.url, data["uuid"]),
             data,
@@ -61,7 +61,7 @@ class RepositoryVariables(BitbucketCloudBase):
 
         return
 
-    def get(self, uuid):
+    def get(self, uuid: str):  # type: ignore[override]
         """
         Returns the pipeline with the uuid in this repository.
 
--- a/atlassian/bitbucket/cloud/repositories/repositoryVariables.py
+++ b/atlassian/bitbucket/cloud/repositories/repositoryVariables.py
@@ -1,5 +1,4 @@
 # coding=utf-8
-from __future__ import annotations
 from ..base import BitbucketCloudBase
 
 
@@ -7,7 +6,7 @@ class RepositoryVariables(BitbucketCloudBase):
     def __init__(self, url, *args, **kwargs):
         super(RepositoryVariables, self).__init__(url, *args, **kwargs)
 
-    def __get_object(self, data) -> RepositoryVariable:
+    def __get_object(self, data) -> "RepositoryVariable":
         return RepositoryVariable(
             self.url_joiner(self.url, data["uuid"]),
             data,
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -6,12 +6,13 @@ import os
 import re
 import time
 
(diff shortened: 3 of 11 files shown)


atlassian/bitbucket/cloud/repositories/repositoryVariables.py
```diff
@@ -61,7 +61,7 @@ def each(self, q=None, sort=None):
 
         return
 
-    def get(self, uuid):
+    def get(self, uuid: str):  # type: ignore[override]
```
bandophahita, line 64: The signature for `get` in `AtlassianRestAPI` doesn't match here.  So we need to override.

requirements.txt
```diff
@@ -1,8 +1,8 @@
-deprecated
+Deprecated
 requests>=2.8.1
 six
 oauthlib
-requests_oauthlib
+requests-oauthlib
```
bandophahita, line 5: not sure why this wasn't included as it's used in the library.

requirements-dev.txt
```diff
@@ -10,6 +10,15 @@ coverage
...
 # used for example confluence attach file
 python-magic
+pylint
+mypy>=0.812
+bandit
+doc8
+types-Deprecated
+types-requests
+types-six
+types-beautifulsoup4
```
bandophahita, line 20: All of the above get used in tox so it only made sense to include them in the dev reqs.

atlassian/rest_client.py
```diff
@@ -480,16 +715,69 @@ def patch(
...
+    def delete(
+        self,
+        path: str,
+        data: dict | str | None = None,
+        headers: dict | None = None,
+        params: dict | None = None,
+        trailing: bool | None = None,
+        absolute: bool = False,
+        advanced_mode: bool = False,
+    ) -> T_resp:
```
bandophahita, line 780: For those unfamiliar with how overloads work this can look confusing.  Essentially this allows mypy to understand that certain arguments change the output type of the method.  Due to some of the of the calls using positional arguments forces us to include additional overrides.   I wish there was an easier way to make this work with mypy.

atlassian/rest_client.py
```diff
@@ -445,15 +680,15 @@ def put(
...
-        path,
-        data=None,
-        headers=None,
-        files=None,
-        trailing=None,
-        params=None,
-        absolute=False,
-        advanced_mode=False,
-    ):
+        path: str,
```
bandophahita, line 683: Not adding the override variations here since `patch` doesn't get used in `jira.py`.

pyproject.toml
```diff
@@ -7,7 +7,6 @@
 [tool.black]
 target-version = ['py36', 'py37', 'py38', 'py39']
 line-length = 120
-include_trailing_comma = false
```
bandophahita, line 10: This isn't a valid configuration parameter.

atlassian/jira.py
```diff
@@ -51,13 +58,16 @@ def _get_paged(
...
 
             while True:
-                response = super(Jira, self).get(
-                    url,
-                    trailing=trailing,
-                    params=params,
-                    data=data,
-                    flags=flags,
-                    absolute=absolute,
+                response = cast(
```
bandophahita, line 61: Overall there is a problem with the current design of `get` in that it could return a dictionary or `None` (same with the other base methods `post`, `put`, and `delete`).  Mypy sees this and complains.  `cast`ing the return of these calls will avoid mypy errors, but this is really just masking the issue.  In theory the code should be doing a check to see if something returned.

examples/jira/jira_review_groups.py
```diff
@@ -66,7 +66,7 @@ def get_inactive_users(groups):
     return inactive_users_list
 
 
-def exclude_inactive_users(groups):
+def exclude_inactive_users(groups: list[dict]):
```
svdimchenko, line 69: It makes sense to add smth using at least TypedDict or even [pydantic model](https://docs.pydantic.dev/latest/concepts/models/#basic-model-usage) as better option
```
from pydantic import BaseModel

class User(BaseModel):
    name: str
    ...

class Group(BaseModel):
    group_name: str
    users: list[User]
    ...

def exclude_inactive_users(groups: list[Group]):
```

gonchik:
@bandophahita once you're ready feel free to update the status of PR

Dockerfile.qa
gonchik, line 1: added, into Dockerfile.qa

examples/jira/jira_review_groups.py
MarcelWilson: I love that idea.  I would recommend that be done in a separate PR since this one solely focuses on adding annotations to existing code.

bandophahita: I love that idea. I would recommend that be done in a separate PR since this one solely focuses on adding annotations to existing code.

atlassian/bitbucket/__init__.py
```diff
@@ -1891,7 +1893,7 @@ def get_pull_requests_participants(
 
     def assign_pull_request_participant_role(
         self, project_key: str, repository_slug: str, pull_request_id: int, role: str, user: str
-    ) -> dict:
+    ) -> Optional[dict]:
```
bandophahita, line 1896: mypy incompatible return triggered once we updated post.

atlassian/confluence.py
```diff
@@ -2798,17 +2800,18 @@ def get_atl_request(url: str):
             parsed_html = BeautifulSoup(response.text, "html.parser")
             # Getting the poll URL to get the export progress status
             try:
-                poll_url = parsed_html.find("meta", {"name": "ajs-pollURI"}).get("content")
+                poll_url = cast("str", parsed_html.find("meta", {"name": "ajs-pollURI"}).get("content"))  # type: ignore[union-attr]
```
bandophahita, line 2803: These got ugly in mypy quickly.  `parsed_html.find` could return `Tag`, `NavigableString`, or `None`; and only `Tag` supports `get()`.   
cast as `str` to avoid mypy error on line 2808.

atlassian/confluence.py
```diff
@@ -2798,17 +2800,18 @@ def get_atl_request(url: str):
...
             except Exception as e:
                 raise ApiError("Problems with getting the poll_url for get_space_export method :", reason=e)
             running_task = True
             while running_task:
                 try:
                     progress_response = self.get(poll_url)
+                    assert progress_response
```
bandophahita, line 2810: Mypy see `get` could return `None`.  Since the following code appears to assume it wasn't `None`, asserting here made the most sense.

atlassian/bitbucket/cloud/repositories/repositoryVariables.py
bandophahita: Since I created the PR it would appear that Codacy also fails a code check
https://app.codacy.com/gh/atlassian-api/atlassian-python-api/pull-requests/1418/issues#issue-997875664eb936e64c346233afbf428c

bandophahita:
I'm about to give up and abandon this PR.  There are just too many hurdles in the codebase to deal with that are outside my purview.  I'll leave it here for now in case someone wants to pick up where I left off.

gonchik:
@bandophahita  could you double check everything is expected from typehints perspective ?

PR #1453, opened 2024-09-14
Continue to support Python 2

madscientist:
Avoid Python 3-only capabilities:
- Remove f-strings (jira.py, confluence.py)
- Remove type specifications (bitbucket)
- Remove trailing commas (bitbucket)

### DIFF
--- a/atlassian/bitbucket/__init__.py
+++ b/atlassian/bitbucket/__init__.py
@@ -2208,13 +2208,13 @@ class Bitbucket(BitbucketBase):
 
     def merge_pull_request(
         self,
-        project_key: str,
-        repository_slug: str,
-        pr_id: int,
-        merge_message: str,
-        close_source_branch: bool = False,
-        merge_strategy: Union[str, MergeStrategy] = MergeStrategy.MERGE_COMMIT,
-        pr_version: Optional[int] = None,
+        project_key,
+        repository_slug,
+        pr_id,
+        merge_message,
+        close_source_branch=False,
+        merge_strategy=MergeStrategy.MERGE_COMMIT,
+        pr_version=None,
     ):
         """
         Merge pull request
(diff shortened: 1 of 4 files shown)


gonchik:
@madscientist  how about to adjust black ? :)

gkowalc:
python2 has been deprecated 4 years ago, I don't see a good reason to keep library  backwards compatibility with python2.x

madscientist:
Well, that's up to the project of course.  But currently the setup for this module says it supports Python 2, but it fails.  This relatively small patch continues to support Python 2.  If the project decides that they no longer want to support Python 2, then they need to modify setup.py to stop saying it does.

Regardless of that, the failures in the online tests above are due to some kind of PPA key failure which has nothing to do with Python 2, and needs to be addressed before I can rerun my tests to verify the patch.

Cheers!

gonchik:
@madscientist  let's continue to support python2

gkowalc:
How about abandoning python2 support in this library starting from the next major  project release (4.x)>. I understand that there are some old codebases which are stuck in python2.7 but committing to support deprecated version of python (by 4 years at this point) will sooner or later cause problems for the maintainers and users of the library. 

Adjusting obvious stuff like f string syntax is easy and came make sense to me, however we shouldn't refrain from using latest  improvements added to  the language in python3.x.  if we want to be fully compatible with python2.7 we should also adjust the code that uses latest features of python, we will be a pain for future maintainers and eventually also for users of the library.

madscientist:
I looked around to try to find a way to re-run the checks in the hopes that whatever the certificate error was, has been resolved, but I don't seem to have permissions to do it (google tells me how to do it but I don't see that option).

Maybe someone else can re-run for me.

gonchik:
@madscientist  thanks

@gkowalc  let's formalize it and prepare an announce for drop py 2.7

madscientist:
Sounds good.  I recommend you arrange to bump the major version when you drop P2 support.  Thanks for pulling!

zffocussss:
python versions before 3.9 are END OF LIFE. Why does this repo continue to support 2.x?

PR #1454, opened 2024-09-15
[Confluence] Add expand option to get_all_pages_by_label

MaksimKovalyov:
Add `expand` in params as `rest/api/content/search` url supports expanding


gonchik:
@MaksimKovalyov Thanks!

PR #1460, opened 2024-09-29
Confluence: Add function get_all_pages_by_space_ids_confluence_cloud

Baarsgaard:
I recently was tasked with bulk exporting metadata about all pages in specific spaces and found out there's a perfect endpoint for that.
It just was not available in the library.

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -720,7 +720,7 @@ class Confluence(AtlassianRestAPI):
         path = "/api/v2/pages"
         params = {}
         if space_ids:
-            params["space-id"] = ','.join(space_ids)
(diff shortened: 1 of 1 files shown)


Baarsgaard:
Formatting has been fixed, on another note, would you accept a PR adding a docker-fmt target to the `Makefile`?:
```make
docker-fmt: docker-qa-build
        docker run --rm -v `pwd`:/atlassian-python-api $(QA_CONTAINER) tox -e black_fmt
```
and `tox.ini`
```ini
[testenv:black_fmt]
basepython = python3
target-version = ["py37"]
skip_install = true
deps = black
commands = black {[base]linting_targets} --exclude __pycache__
```

PR #1476, opened 2024-11-26
Improved download_attachments_from_page: added filename filter for si…

AImenes:
Here’s an improved and generalized version of your pull request comment:

---

#### **`download_attachments_from_page` Method in `confluence.py`**
- **Backward Compatibility**: The existing usage of this method remains unchanged, and can be used as previous for downloading all attachments on page.
- **New Features**:
  1. **Download by Filename**: Added the `filename` parameter to enable downloading a specific file by its name. This avoids the need to process all attachments when only one file is required.
  2. **In-Memory Downloads**: Introduced the `to_memory` parameter. When set to `True`, files are downloaded as `BytesIO` objects instead of being written to disk. This is particularly useful when files are passed directly into other pipelines, such as `pandas`, without intermediate storage.

#### **Example Use Case**:
```python
from atlassian import Confluence
import pandas as pd

# Initialize Confluence client
confluence = Confluence(
    url=base_url,
    token=open(token_path, 'r').read().strip(),
)

# Retrieve the page ID
page_id = confluence.get_page_id("some_space", "some_page")

# Download a specific file directly into memory
file_name = "excel_file.xlsx"
file_data = confluence.download_attachments_from_page(page_id=page_id, filename=file_name, to_memory=True)

# Process the file using pandas
excel_data = pd.read_excel(io=file_data.get(file_name))

# Output: DataFrame ready for use in further analysis
print(excel_data)
```

---

### **Other Changes**:
- **PEP-8 Compliance**: Reorganized imports in `confluence.py`:
  - Alphabetically ordered standard library, third-party, and local imports.
---

### DIFF
--- a/atlassian/confluence.py
+++ b/atlassian/confluence.py
@@ -1,15 +1,25 @@
 # coding=utf-8
+import io
+import json
 import logging
 import os
-import time
-import json
 import re
-from requests import HTTPError
-import requests
-from deprecated import deprecated
+import time
+
 from bs4 import BeautifulSoup
+from deprecated import deprecated
+import requests
+from requests import HTTPError
+
 from atlassian import utils
-from .errors import ApiError, ApiNotFoundError, ApiPermissionError, ApiValueError, ApiConflictError, ApiNotAcceptable
+from .errors import (
+    ApiConflictError,
+    ApiError,
+    ApiNotAcceptable,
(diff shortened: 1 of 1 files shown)


gonchik:
@AImenes  Thanks, let me adjust for f string

Issue #1289, opened 2024-01-09
Rate limits applied on instance - our code borked


GanizaniSitara:
Our infra team have switched on rate limits on our instance (no issues with that, right thing to do), but consequently our publishing code is borked. What do people do? Monkey patch the Atlassian library with a rate limit handler? Are there any 'proper' approaches to this? Thanks!

gonchik:
@GanizaniSitara  we can try to add retry for 429 error code

GanizaniSitara:
That would be amazing. There's a recommended back-off pseudo code published by Atlassian here https://developer.atlassian.com/cloud/jira/platform/rate-limiting/  I started to implement it only to realize I'm essentially rewriting your wrapper ...

GanizaniSitara:
Hi @mmahacek, yes it seems that way, I've only been made aware of the "Retry-After" headers by our support team in the last couple of days and haven't explored them yet. I had been thinking about he behaviour of the library - I think I would want to log a warning (as opposed to error) for the 429s and retry? But happy to be persuaded otherwise. Also there seem to be two types of limits with different behaviour which I saw after putting in my rudimentary rate limit wrapper. One is "rate limit" and the other is "too many messages" - that one seems to return a body with the response. I guess that might depend on how the instance is configured. Just something to flag for anyone testing this.

PR #1486, opened 2025-01-09
Support HTTP 429 Retry-After

janbrummer:
Check for Retry-After and retry after that period.

Fixes: https://github.com/atlassian-api/atlassian-python-api/issues/1289


GanizaniSitara:
Hi Jan, thank you for this, much appreciated! Will look out for it in the next version.

gonchik:
uploading a new release

PR #1491, opened 2025-01-20
[Jira] Fix Jira archive project REST API method from POST to PUT

SuperWonjune:
https://docs.atlassian.com/software/jira/docs/api/REST/9.14.0/#api/2/project-archiveProject

In the latest Jira LTS versions (9.12, 9.4), a 405 error occurs with the "HTTP request Wrong method" when using the Archive Project function.

Upon investigation, it was found that the method was incorrectly set to POST. This will be corrected to use PUT instead.

### DIFF
--- a/atlassian/jira.py
+++ b/atlassian/jira.py
@@ -2508,7 +2508,7 @@ class Jira(AtlassianRestAPI):
         """
         base_url = self.resource_url("project")
         url = "{base_url}/{key}/archive".format(base_url=base_url, key=key)
-        return self.post(url)
+        return self.put(url)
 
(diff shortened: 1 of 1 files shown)


gonchik:
Thank for that detection, 
indeed for 8.13 the same
https://docs.atlassian.com/software/jira/docs/api/REST/8.13.0/#api/2/project-archiveProject
https://docs.atlassian.com/software/jira/docs/api/REST/8.1.0/#api/2/project-archiveProject
https://docs.atlassian.com/software/jira/docs/api/REST/7.12.0/#api/2/project-archiveProject

PR #1506, opened 2025-03-08
Add pagelen to bitbucket cloud repositories and branches each functions

aaronsewall:
Didn't see another way to change this so I added this in the two places I tried it, seems to work with pagelen=100.

### DIFF
--- a/atlassian/bitbucket/cloud/repositories/refs.py
+++ b/atlassian/bitbucket/cloud/repositories/refs.py
@@ -51,7 +51,7 @@ class Refs(BitbucketCloudBase):
         if q is not None:
             params["q"] = q
         if pagelen is not None:
-            params["pagelen"] = q
+            params["pagelen"] = pagelin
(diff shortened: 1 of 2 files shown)


atlassian/bitbucket/cloud/repositories/refs.py
```diff
@@ -32,22 +32,26 @@ def create(
...
+        
         :return: A generator for the Ref objects
         """
         params = {}
         if sort is not None:
             params["sort"] = sort
         if q is not None:
             params["q"] = q
+        if pagelen is not None:
+            params["pagelen"] = q
```
Viktxrrr, line 54: I think there should be `pagelen` instead of `q` here

Viktxrrr, line 54: I think there should be `pagelen` instead of `q` here

Viktxrrr (changes requested):

PR #1510, opened 2025-03-18
fix: [Confluence] download_attachments_from_page: the return value from the response is bytes.

mgomes-d:
- Adjusted the download logic to process the file content directly as bytes.


meyerjo1SICKAG:
I think one additinally should add `, absolute=True` to the request as in line 1471 https://github.com/mgomes-d/atlassian-python-api/blob/cca91e9dc942c57d61027089ef656fbc7366a071/atlassian/confluence.py#L1471 the self.url is prepended to the link. Alternatively, one could just use the download link and let the requests library handle that.

PR #1522, opened 2025-04-03
Confluence api v2 and Jira api v3 added

batzel:
I've added confluence v2 and jira v3, for the newer Atlassian Cloud servers. Unit tests and integration tests against live servers, flake8 and black to clean up the code, and I'm using it now with mcp-atlassian.


gonchik:
@batzel  @batzel-upenn  why do you add into 1 merge request 2 different wrapers ?
I see many other files are changed. 
could you make changes more atomic and small bunches please ? 
Code review is starting to be very complicated, as I previous code I did not read fully yet

gonchik (commented):
that's not needed

batzel:
Sorry, I'm a bit new at working with public repos. My main goal was getting the code working for me. I can separate things out into several requests if you want it that way. Can you suggest how you'd like me to break it into sections? I don't mean to cause you trouble.

gonchik:
Both modules are implemented

PR #1523, opened 2025-04-06
Confluence v2 implementation

gonchik:
only Confluence v2


test_search.py
```diff
@@ -0,0 +1,87 @@
...
+# Check if environment variables are loaded
+if not all([CONFLUENCE_URL, CONFLUENCE_USERNAME, CONFLUENCE_API_TOKEN, SPACE_KEY]):
+    print("Error: Missing environment variables. Please create a .env file with the required variables.")
+    exit(1)
+
+# Test with just a query
+print("Test 1: Search with simple query")
+query = "test"
+response = requests.get(
+    f"{CONFLUENCE_URL}/wiki/api/v2/search",
```
qvalentin, line 26: I can not find this endpoint in the docs (https://developer.atlassian.com/cloud/confluence/rest/v2/intro/#about) and it also does not work for my instance.

gonchik: Let me check it

gonchik: Indeed

justinfx:
Can someone comment on the the status of this ticket, vs seemingly similar work in https://github.com/atlassian-api/atlassian-python-api/pull/1522?

Is this, or the other MR, required for this library to actually support Confluence Cloud servers that are deployed with the v2 REST API? Without this work, is this library not usable in that context?

loganmzz:
What is the current status? Is it possible to use scoped token (mandatory for service accounts)?

gonchik:
Implemented

PR #1532, opened 2025-05-05
Make beautifulsoup4 an Optional Dependency

Lukas-solar:
#### Summary

This PR moves `beautifulsoup4` to an **optional dependency**, installable via the `html` extra:

```bash
pip install atlassian-python-api[html]
```

#### Rationale

* **Limited Usage Scope**: `beautifulsoup4` is only required for HTML parsing in a small subset of modules. The majority of the package's functionality does not depend on it.
* **Reduces Installation Overhead**: By not enforcing installation of unused dependencies, we streamline the installation process for users who don't need HTML parsing.
* **Improves Compatibility**: Some environments (e.g., minimal Docker images or serverless functions) may benefit from fewer dependencies, faster builds, and smaller footprints.
* **Design Philosophy**: Since HTML parsing can be seen as a less "clean" or ideal approach (often used as a workaround when APIs or structured data aren't available), it's appropriate to isolate this functionality and not impose it on all users by default.

#### Usage Note

If a user requires HTML-related features, they can explicitly install the optional dependency.
Attempting to use the HTML-parsing functionality without installing `beautifulsoup4` will raise an informative `ImportError`.


gonchik:
@Lukas-solar  Thank you! Indeed that very good step. 

Let's  try to deploy it

PR #1543, opened 2025-06-06
Visualization for the atlassian-python-api codebase

ivanmilevtues:
Generated documentation to help new developers to get up to speed with the atlassian-python-api.

Hey all. Me and a friend have noticed that it is quite hard to get to know new codebases, so we are working on something which can help people get up to speed and thus reducing the load for both maintainers and new devs. We believe that it is always best to start with nice high-level visualization and then dig down into the interesting/relevant parts of the code i.e. first task/something one wants to contribute to.

This said, we generated diagrams leveraging static analysis and LLM's.  I generated one for the atlassian-python-api project.
You can see how the added change renders in the github markdown here - https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/atlassian-python-api/on_boarding.md

Would love to hear your opinion on the diagrams. Further I know atlassian is a big company so I suppose there is some process already in place, can you share some details on what it looks like, do you think such diagrams can fit into it? Further what is your opinion on diagram driven on-boarding?

Any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Atlassian Platform & Marketplace Clients.md
+++ b/.codeboarding/Atlassian Platform & Marketplace Clients.md
@@ -1,0 +1,234 @@
+```mermaid
+graph LR
+    AtlassianRestAPI["AtlassianRestAPI"]
+    Crowd_API_Client["Crowd API Client"]
+    StatusPage_API_Client["StatusPage API Client"]
+    MarketPlace_API_Client["MarketPlace API Client"]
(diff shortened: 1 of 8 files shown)


gonchik:
@ivanmilevtues  Let me merge. 
Actually it will be quite simple for that wrapper. 

more dynamics will be nice :)

Issue #1580, opened 2025-08-29
The field id is missing in the error message


zhan9san:
https://github.com/atlassian-api/atlassian-python-api/blob/df81dc97236d3a502dd4ea371cbe9adf43922ce6/atlassian/rest_client.py#L988-L989

```log
  File "/usr/local/lib/python3.11/site-packages/atlassian/jira.py", line 1861, in set_issue_status
    return self.post(url, data=data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/atlassian/rest_client.py", line 388, in post
    response = self.request(
               ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/atlassian/rest_client.py", line 312, in request
    self.raise_for_status(response)
  File "/usr/local/lib/python3.11/site-packages/atlassian/rest_client.py", line 551, in raise_for_status
    raise HTTPError(error_msg, response=response)
requests.exceptions.HTTPError: This field is required
```

I find the dict is supported in `HTTPError`. Is it necessary to convert it to a string?

```python
from requests import HTTPError

errors = {
    "duedate": "Field 'duedate' cannot be set. It is not on the appropriate screen, or unknown.",
    "assignee": "Field 'assignee' cannot be set. It is not on the appropriate screen, or unknown.",
}

raise HTTPError(errors)

```

Output
```
Traceback (most recent call last):
    raise HTTPError(errors)
requests.exceptions.HTTPError: {'duedate': "Field 'duedate' cannot be set. It is not on the appropriate screen, or unknown.", 'assignee': "Field 'assignee' cannot be set. It is not on the appropriate screen, or unknown."}
```

gonchik:
This converts the dict of error messages into a human-readable string that appears in the exception message instead of a raw dict.

If you find use case of improvement, feel free to improve by PR. 
I am happy to check and merge it

PR #1583, opened 2025-08-31
Error exceptions

gonchik:
#1580


zhan9san:
@gonchik 

thanks for your help.

the field id as keys is lost in line 989 as it uses errors.values().

                        error_msg_list.extend(list(errors.values()))

would you help update this line?

PR #1587, opened 2025-09-12
Jira: Add handling for errorMessages field in JSM responses

cmpaul:
We are occasionally receiving errors from JSM requests like this:

```json
{
  errorMessages: [
    "Field Assignee is required."
  ],
  errors: {}
}
```

This results in a `KeyError: 'errorMessage'` exception.


gonchik:
@cmpaul thanks

PR #1624, opened 2026-03-13
Run black formatter - outdated version for Python 3.9

jerubball:
Recent merges had checks failing again.

I have run the black formatter, and also commented out unused import to have flake8 pass.

As pointed out in #1619, latest version of `black==26.3.1` does not support Python 3.9.

We have two options that I can think of.

- One is to drop the Python 3.9 support, because it reached end of life on October 2025.

- Other is to set the black version in `requirements-dev.txt` file to `black==25.11.0`.

This PR only runs formatter, and does not implement the solutions above.


jerubball:
Closing as #1625 includes formatter fixes

PR #1654, opened 2026-08-17
Preserve Confluence Server REST URL compatibility

hkwi:
## Summary

The ConfluenceServer class in 5.0.0 changed the client to append the configured API root and version to its base URL. Existing calls that already pass paths such as `rest/api/content/...` therefore construct URLs such as `/rest/api/1.0/rest/api/content/...`.

This breaks the established Confluence Server REST API URL used by on-premises installations. It also incorrectly places non-REST browser endpoints, such as PDF and Word exports, below the REST API root.

## Changes

- Keep the Confluence site/context URL as the `ConfluenceServer` base URL, matching 4.x behavior.
- Prefix only unrooted Server REST resources at request time.
- Leave legacy `rest/api/...` paths, experimental REST paths, attachment download links, and browser/UI paths unchanged.
- Apply explicitly configured `api_root` and `api_version` values to unrooted REST resources without duplicating rooted paths.
- Preserve the existing method-level path expectations instead of rewriting them to fit the new URL layout.
- Add final HTTP URL regression coverage for relative REST resources, 4.x-style `post`/`put`/`delete` paths, explicit API versions, and PDF/Word export paths.

This keeps the Cloud/Server class split introduced in 5.x while restoring compatibility for existing Confluence Server clients.

## Testing

- Focused Confluence tests: 168 passed, 6 skipped.
- Full test suite: 1027 passed, 10 skipped.
- Black check: passed for both changed files.

### DIFF
--- a/atlassian/confluence/server/__init__.py
+++ b/atlassian/confluence/server/__init__.py
@@ -42,6 +42,25 @@ class Server(ConfluenceServerBase):
     Confluence Server REST API wrapper
     """
 
+    _api_resources = frozenset(
+        {
+            "accessmode",
+            "admin",
+            "audit",
+            "content",
+            "contentbody",
+            "group",
+            "health",
+            "longtask",
+            "metadata",
+            "reindex",
+            "search",
+            "space",
+            "template",
+            "user",
+        }
+    )
+
     content_types = {
         ".gif": "image/gif",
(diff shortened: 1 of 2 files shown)


gonchik:
@hkwi  thanks

PR #1659, opened 2026-08-20
fix incorrect keyword in requests

Tander:
According to https://developer.atlassian.com/cloud/assets/rest/api-group-object/#api-object-navlist-iql-post it should be results and not result


gonchik:
@Tander  what's happened?

